PR #2, opened 2017-08-03
[Mod-43][feature] Add base for provider moderation page

sheriefvt:
# Ticket:
https://openscience.atlassian.net/browse/MOD-43

# Purpose:
Add the base for the provider moderation page (tabs, provider name, and breadcrumbs) to an ember component. The component is designed to allow adding new tabs and their corresponding routes in the futue. Breadcumbs is built using ember-crumbly and ui-navigator-service.

image


laurenbarker:
@sheriefvt, I added a spec document and the breadcrumbs separator image to the Jira ticket; sorry I left those out to begin with. Just a few nitpicks since you didn't have the spec: 
* The separator image should be used between breadcrumbs instead of ">". If that is too difficult then we can leave it out for MVP.
* Only the current page of the breadcrumbs should be bolded and not clickable. 
* Setting shouldn't be in the breadcrumbs since the user will think they are on the same page when they switch tabs. 
* The font sizes (they can be seen in the spec) look a little off from the mockup.

All minor changes though, looks good!

app/components/moderation-base.js
```diff
@@ -0,0 +1,35 @@
+import Ember from 'ember';
+/**
+ * The base for the provider moderation page (tabs, provider name, and breadcrumbs).
+ *
+ * Sample usage:
+ * ```handlebars
+ * {{moderation-base}}
```
aaxelb: Tiny nitpick: `{{moderation-base`, without the closing `}}`

ember-cli-build.js
```diff
@@ -40,7 +40,9 @@ module.exports = function(defaults) {
             [useCdn ? 'jquery.js' : '']: false,
         },
         'ember-bootstrap': {
-            importBootstrapCSS: false
+          importBootstrapCSS: false,
```
aaxelb: 4-space indentation

app/routes/provider.js
```diff
@@ -0,0 +1,44 @@
...
+ * @class Provider Route Handler
+ */
+export default Ember.Route.extend({
+    theme: Ember.inject.service(),
+
+    beforeModel(transition) {
+        const {slug = ''} = transition.params.provider;
+        const slugLower = slug.toLowerCase();
+
+        if (providerIds.includes(slugLower)) {
```
aaxelb: I would rather not have a hard-coded list of providers, so we don't have to update code whenever new providers are added. This looks like it was copied from preprints to support resolving GUIDs... do we need a `beforeModel` at all? Seems like this route should just have a `model` method that gets the provider by id.

app/components/moderation-base.js
```diff
@@ -0,0 +1,35 @@
+import Ember from 'ember';
```
aaxelb: Please generate components with `--pod`, so this would live at `app/components/moderation-base/component.js`, and this component's template at `app/components/moderation-base/template.hbs`.

aaxelb (commented):
Looks good, just a couple things. ⛵️

app/routes/provider.js
sheriefvt: @aaxelb I removed some of the logic and kept one to check for provider id. I added also a **todo** note to remove in future

sheriefvt:
@laurenbarker @aaxelb thanks for your feedback..looks much better now. 
I added also a page-not-found page. Font family might look slightly different... this will be fixed when other PR #3  for **reviews landing page** is merged.

app/components/moderation-base/template.hbs
```diff
@@ -0,0 +1,38 @@
+<div class="dark-header"> {{!HEADER}}
+    <div class="container"> {{!CONTAINER}}
+        <div class="row">{{!HEADER ROW}}
+            <div class="text-left p-l-lg col-md-12">
+                <p>
+                    {{#each breadCrumbs as |breadCrumb|}}
```
chrisseto: Would it be better to use an existing bread crumbs library? Do you know if we already use one?

app/routes/provider.js
```diff
@@ -0,0 +1,33 @@
...
+export default Ember.Route.extend({
+    theme: Ember.inject.service(),
+
+    // Todo: Replace the use of hardcoded preprint provider list with an API request.
+
+    beforeModel(transition) {
+        const {slug = ''} = transition.params.provider;
+        const slugLower = slug.toLowerCase();
+        if (providerIds.includes(slugLower)) {
+            if (slugLower !== slug) {
```
chrisseto: You should use ember modify the URL here, maybe something [built into route](https://emberjs.com/api/ember/2.14/classes/Ember.Route)?

app/styles/app.scss
```diff
@@ -20,16 +20,30 @@
   list-style-type: none;
 }
 
-
-
 .reviews-brand {
-  background: url('img/reviews-logo.png') top center no-repeat;
-  height: 70px;
-  background-size: contain;
+  background: url('img/reviews-logo-beta.png') top center no-repeat;
```
chrisseto: Are we officially labeling this beta?

config/environment.js
```diff
@@ -34,19 +34,64 @@ module.exports = function(environment) {
...
-            // OSF must be the first provider
-            {
-                id: 'osf',
-                logoSharing: {
-                    path: '/assets/img/provider_logos/osf-dark.png',
-                    type: 'image/png',
-                    width: 363,
-                    height: 242
+            providers: [
+                // OSF must be the first provider
```
chrisseto: Why is this the case?

config/environment.js
```diff
@@ -34,19 +34,64 @@ module.exports = function(environment) {
         },
         Reviews: {
             defaultProvider: 'osf',
```
chrisseto: I don't know if it makes sense to have a default provider for reviews

ember-cli-build.js
```diff
@@ -40,7 +40,9 @@ module.exports = function(defaults) {
             [useCdn ? 'jquery.js' : '']: false,
         },
         'ember-bootstrap': {
-            importBootstrapCSS: false
+            importBootstrapCSS: false,
```
chrisseto, line 43: Fix the spacing here

package.json
```diff
@@ -37,22 +39,25 @@
     "ember-cli-sass": "5.3.1",
     "ember-cli-shims": "1.0.2",
     "ember-cli-sri": "2.1.0",
+    "ember-cli-string-helpers": "1.4.0",
     "ember-cli-template-lint": "0.4.12",
     "ember-cli-test-loader": "1.1.0",
     "ember-cli-uglify": "1.2.0",
+    "ember-crumbly": "1.0.7",
```
chrisseto: I don't see this being used anywhere

package.json
```diff
@@ -15,7 +15,9 @@
     "test": "ember test"
   },
   "devDependencies": {
+    "@centerforopenscience/ember-osf": "0.7.1",
     "autoprefixer": "6.3.7",
+    "bootstrap-sass": "^3.3.7",
```
chrisseto: What is this being added for?

package.json
```diff
@@ -37,22 +39,25 @@
     "ember-cli-sass": "5.3.1",
     "ember-cli-shims": "1.0.2",
     "ember-cli-sri": "2.1.0",
+    "ember-cli-string-helpers": "1.4.0",
```
chrisseto: I don't think this is worth installing if we're only using it for capitalize. @laurenbarker has some cool tricks up her sleeve to capitalize things with CSS

package.json
```diff
@@ -37,22 +39,25 @@
...
     "ember-power-select": "1.0.0-beta.7",
     "ember-resolver": "2.1.0",
+    "ember-responsive-tabs": "0.0.10",
     "ember-sanitize": "2.0.2",
     "ember-sortable": "1.8.2",
     "eslint": "3.18.0",
     "loader.js": "4.5.1",
-    "postcss": "5.1.0"
+    "postcss": "5.1.0",
+    "ui-navigator-service": "0.3.11"
```
chrisseto: If crumbly is used, do we actually need this?

package.json
```diff
@@ -37,22 +39,25 @@
...
+    "ember-crumbly": "1.0.7",
     "ember-data": "2.7.0",
     "ember-export-application-global": "1.0.5",
     "ember-font-awesome": "2.1.1",
     "ember-get-config": "0.0.4",
     "ember-load-initializers": "0.5.1",
-    "@centerforopenscience/ember-osf": "0.7.1",
     "ember-power-select": "1.0.0-beta.7",
     "ember-resolver": "2.1.0",
+    "ember-responsive-tabs": "0.0.10",
```
chrisseto: It looks like this is not being used either?

chrisseto (changes requested):

app/styles/app.scss
laurenbarker: @chrisseto, that is what I suggested for the first release (it was approved). The first release will most likely be private beta to only the current preprint providers. I imagine beta would come off once OSF admin aren't required to give admin the permissions to setup moderation.

I haven't heard anything against labelling it as beta, have you?

app/components/moderation-base/template.hbs
sheriefvt: I don't think it is used in preprints

package.json
sheriefvt: i think dependency by some ember-osf components

package.json
sheriefvt: yes we use both

app/styles/app.scss
chrisseto: Nope, just double checking.

ember-cli-build.js
sheriefvt: because of line 4 in [ routes/provider.js](https://github.com/CenterForOpenScience/ember-osf-reviews/pull/2/files/1908a8109788fb5618db924e1d2e61798c2aeb47#diff-b4af49876da16fe17ebc9b388cd01841)
``
const providers = config.Reviews.providers.slice(1);
``

config/environment.js
sheriefvt: because of line 4 in [ routes/provider.js](https://github.com/CenterForOpenScience/ember-osf-reviews/pull/2/files/1908a8109788fb5618db924e1d2e61798c2aeb47#diff-b4af49876da16fe17ebc9b388cd01841)
``
const providers = config.Reviews.providers.slice(1);
``

package.json
sheriefvt: i will remove crumbly

app/styles/app.scss
```diff
@@ -175,38 +158,60 @@
...
+  padding-left: 20px;
+  padding-right: 20px;
 }
 
 #tabTitle .nav > li > a:hover {
-  color: black;
+  color: #333333;
+}
+
+@media (max-device-width: 400px) {
```
chrisseto: Use [bootstrap's breakpoint variables](https://getbootstrap.com/docs/3.3/css/#grid-media-queries)

app/templates/provider/moderation.hbs
```diff
@@ -0,0 +1,2 @@
+{{outlet}}
+{{moderation-base provider-title='SocArXiv' provider-id='socarxiv' unread-count=10 active='Moderation'}}
```
chrisseto: Hard coded values?

app/templates/provider/settings.hbs
```diff
@@ -0,0 +1,2 @@
+{{outlet}}
+{{moderation-base provider-title='SocArXiv' provider-id='socarxiv' unread-count=10 active='Settings'}}
```
chrisseto: Hard coded values?

app/templates/provider/moderation.hbs
chrisseto: The `moderation-base` component should be in `app/templates/providers.hbs`

config/environment.js
```diff
@@ -34,19 +34,64 @@ module.exports = function(environment) {
...
-        providers: [
-            // OSF must be the first provider
-            {
-                id: 'osf',
-                logoSharing: {
-                    path: '/assets/img/provider_logos/osf-dark.png',
-                    type: 'image/png',
-                    width: 363,
-                    height: 242
+            providers: [
```
chrisseto: I believe we determined this could be removed.

app/routes/dashboard.js
```diff
@@ -0,0 +1,4 @@
+import Ember from 'ember';
```
chrisseto: There is no need to define empty routes, ember will do this implicitly.

app/routes/page-not-found.js
```diff
@@ -0,0 +1,4 @@
+import Ember from 'ember';
```
chrisseto: There is no need to define empty routes, ember will do this implicitly.

chrisseto (changes requested):

config/environment.js
sheriefvt: this is addressed in PR #8

app/templates/provider/settings.hbs
sheriefvt: this is addressed in PR #8

app/templates/provider/moderation.hbs
sheriefvt: this is addressed in PR #8

sheriefvt: this is addressed in PR #8

app/components/error-page/template.hbs
```diff
@@ -0,0 +1,18 @@
+<div class="container">
```
chrisseto: This should use the translation file

app/styles/app.scss
```diff
@@ -128,30 +114,27 @@
   padding-top: 30px;
   padding-bottom: 30px;
   border-bottom: 1px solid #dddddd;
-  color: #333333;
+  color: black;
```
chrisseto: @laurenbarker, Do you happen to have hex for the darkest acceptable color?
@sheriefvt Black is generally too much contrast. A *very* dark grey is more appealing to the eye.

app/styles/app.scss
chrisseto: @sheriefvt

app/routes/provider.js
```diff
@@ -0,0 +1,32 @@
...
+ * @submodule routes
+ */
+
+/**
+ * @class Provider Route Handler
+ */
+export default Ember.Route.extend({
+    theme: Ember.inject.service(),
+
+    beforeModel(transition) {
```
chrisseto: This should all be handled in the `model` hook.

app/routes/provider.js
```diff
@@ -0,0 +1,32 @@
...
+        const slugLower = slug.toLowerCase();
+        const upstreamPromise = this._super();
+        this.store.find('preprint-provider', slugLower).then((provider) =>{
+            this.set('theme.id', provider.id);
+        }).catch(() =>{
+            this.replaceWith('page-not-found');
+        });
+        return Ember.RSVP.resolve(upstreamPromise)
+            .then(function() {
+                return new Ember.RSVP.Promise(function(resolve) {
```
chrisseto: You shouldn't need a time out here.

app/components/moderation-base/component.js
```diff
@@ -0,0 +1,44 @@
...
+ *    unread-count=10
+ *    active='Settings'
+ * }}
+ * ```
+ * @class moderation-base
+ **/
+export default Ember.Component.extend({
+    theme: Ember.inject.service(),
+    tabs: [
+        { name: 'Moderation', route: 'provider.moderation'},
```
chrisseto: I18n?

chrisseto (changes requested):

app/styles/app.scss
laurenbarker: It's `#333333`.

ember-cli-build.js
```diff
@@ -40,7 +40,9 @@ module.exports = function(defaults) {
...
+            'bootstrapVersion': 3,
+            'importBootstrapFont': true
```
chrisseto: This should be `false`, we're not using glyphicons for anything

app/styles/app.scss
chrisseto:
```scss
@import 'bootstrap/variables';

@media (max-device-width: $screen-sm/xs/md/etc-min) {
}
```

chrisseto:
Just 2 minor things, ready to go after that 👍

sheriefvt:
@chrisseto done with changes.. thanks for feedback 👍

chrisseto (approved):

PR #3, opened 2017-08-03
[MOD-46][feature] Add osf-reviews landing page

sheriefvt:
# Ticket:
https://openscience.atlassian.net/browse/MOD-46

# Purpose:
Add landing page for OSFReviews. 

# Steps:
* Update index.hbs
* Add needed images
* Use icons for bullets and adjust styling

**Note**: the "Start moderation" button will be in a separate ticket.
## screenshot
image


app/templates/index.hbs
```diff
@@ -5,12 +5,59 @@
...
-
         </div> {{!END CONTAINER}}
     </div> {{!END HEADER}}
+    <div class="reviews-middle"> {{!MIDDLE}}
+        <div class="container"> {{!CONTAINER}}
+            <div class="row col-md-6">
+                <div class="reviews-middle-features"></div>
+            </div>
+            <div class="col-md-6">
+                <h2>Moderate Your Collection</h2>
```
aaxelb: This should be "Moderate your collection", like in the mockup

aaxelb (commented):
Some visual things:
- The styling for the "Moderate Your Collection" and "Want to start a moderated service?" blocks don't match the mockup. Try to get the fonts, weights, and spacing to line up with what was approved.
- The key and globe icons under "Choose Your Workflow" should be aligned with the top of the text, like in the mockup. Also, match the font styling.
- The logo is bigger than in the mockup, but is closer to the logo size on preprints/registries. We should probably try to match the other services as close as we can.
- The flow charts aren't offset like the mockup -- not sure how much that one matters.
- The nav bar doesn't look like the one I see on other services -- there's too much padding and the font size is off. Are we using a different version of ember-osf, or are some of our styles interfering?

app/templates/index.hbs
sheriefvt: @aaxelb thanks for your feedback, I will address your requests. The navbar is using the OSF styling we imported. It is same as one in preprints and used from ember-osf. There is another version of nav-bar used for preprint providers page. We are using the same version of ember-osf.

laurenbarker:
@sheriefvt, I added the spec to the Jira ticket so that font size, line height, and spacing are available. There should probably be a max width on the text boxes to prevent them from getting too wide to read easily.

sheriefvt:
@laurenbarker @aaxelb thanks for feedback updated with requested changes

chrisseto (approved):

PR #5, opened 2017-08-22
[MOD-62][Feature] Install i18n and add translation file

laurenbarker:
## Purpose
Add translation file.

## Changes
* Install i18n
* Add common words to translation file
* Define missing values in theme.js

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -10,4 +10,28 @@ import OSFAgnosticAuthRouteMixin from 'ember-osf/mixins/osf-agnostic-auth-route'
 /**
  * @class Application Route Handler
  */
-export default Ember.Route.extend(OSFAgnosticAuthRouteMixin);
+export default Ember.Route.extend(OSFAgnosticAuthRouteMixin, {
+    i18n: Ember.inject.service(),
+    afterModel: function() {
+        const availableLocales = this.get('i18n.locales').toArray();
+        let locale;
+
+        // Works in Chrome and Firefox (editable in settings)
+        if (navigator.languages && navigator.languages.length) {
+            for (let lang of navigator.languages) {
+                if (availableLocales.includes(lang)) {
+                    locale = lang;
+                    break;
+                }
+            }
+        }
+        // Backup for Safari (uses system settings)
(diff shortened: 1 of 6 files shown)


app/routes/application.js
```diff
@@ -10,4 +10,28 @@ import OSFAgnosticAuthRouteMixin from 'ember-osf/mixins/osf-agnostic-auth-route'
...
+        if (navigator.languages && navigator.languages.length) {
+            for (let lang of navigator.languages) {
+                if (availableLocales.includes(lang)) {
+                    locale = lang;
+                    break;
+                }
+            }
+        }
+        // Backup for Safari (uses system settings)
+        else if (navigator.language && availableLocales.includes(navigator.language)) {
```
chrisseto: Our style guide prefers `} else if (...) {`. Putting the comment beneath the if shouldn't be too confusing

config/environment.js
```diff
@@ -34,6 +34,9 @@ module.exports = function(environment) {
         },
         Reviews: {
             defaultProvider: 'osf',
+            i18n: {
```
chrisseto: How does i18n know to read out of the `Reviews` key?

app/services/theme.js
```diff
@@ -32,4 +32,17 @@ export default Ember.Service.extend({
         return `${config.OSF.url}register?${query}`;
     }),
 
+    // The provider object
+    provider: Ember.computed('id', function() {
```
chrisseto, line 36: What do we need this for?

chrisseto (changes requested):
2 questions and a nitpick

app/services/theme.js
laurenbarker: ember-osf errors if provider isn't defined :/

app/routes/application.js
laurenbarker: Sure. PSA: the code is directly copied from preprints.

config/environment.js
laurenbarker: Oops this addition shouldn't be here, it's already in the file.

laurenbarker:
@chrisseto, changes addressed 💸

chrisseto (approved):

PR #7, opened 2017-08-23
[WIP][MOD-45][Feature] Dashboard

aaxelb:


app/styles/_variables.scss
```diff
@@ -5,3 +5,18 @@ $reviews-title-hover: #D5E6ED;
 $gradient-percentage: 15%;
 $nav-bar-height: 25px;
 $carousel-control-width: 10%;
+
+$color-action-submit: #CABC14;
+$color-action-accept: #5CB85C;
+$color-action-reject: #D9534F;
+$color-action-edit-comment: #337ab7;
+
+$color-text-black: #151515;
```
laurenbarker: I'm pretty sure `#333333` is the darkest text color we should be using. Was `#151515` in the spec?

aaxelb: Yep, for the bold/large title of each review log on the dashboard, e.g. "Karl Marx accepted a preprint". I'll redefine black to `#333333` and use it more consistently.

laurenbarker: Ah, my bad. Thanks!

app/controllers/index.js
```diff
@@ -0,0 +1,10 @@
+import Ember from 'ember';
+
+export default Ember.Controller.extend({
+
+    showDashboard: Ember.computed('model.[]', function() {
+        const providers = this.get('model');
+        return providers.any((p) => p.get('reviews_workflow') !== 'none');
```
chrisseto: Should this be null now?

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "reviews",
   "version": "0.0.0",
   "description": "Center for Open Science Reviews Service",
-  "license": "MIT",
+  "license": "Apache-2.0",
```
chrisseto, line 5: +10

app/routes/index.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+export default Ember.Route.extend({
+    store: Ember.inject.service(),
+    currentUser: Ember.inject.service(),
+
+    model() {
+        return this.get('store').query('preprint-provider', {
+            'filter[permissions]': 'view_review_logs'
```
chrisseto, line 9: Is there a way to do `OR`s in the API?

aaxelb: Not that I can see... both `filter[permissions]=view_review_logs,set_up_moderation` and `filter[permissions]=view_review_logs&filter[permissions]=set_up_moderation` behave like an `AND`. Since admin permissions are a superset of moderator permissions, though, we could just query for `view_review_logs` and filter by `set_up_moderation` in the browser...

aaxelb: Oh wait, I wrote the comma part, it's not in every filter... I made it `AND` because it seemed more consistent with the rest of the API. But looking again, the only other place a comma-separated list happens is for a list of guids, where it's an `OR`. I guess I should make `filter[permissions]=view_review_logs,set_up_moderation` be an `OR`

chrisseto:
Merged via https://github.com/CenterForOpenScience/ember-osf-reviews/commit/591719d65802eb7528db65727af65258383fe192

PR #8, opened 2017-08-23
[Mod-44][feature] Add Moderation list

sheriefvt:
# Ticket:
https://openscience.atlassian.net/browse/MOD-44

# Purpose:
Create the moderation list view for moderators.

# Steps:
* Add moderation list component
* Add accepted, rejected, and pending components
* Add pager service and use pagination-pager add-on
* Update css styling
* Use deprecation workflow to hide warnings
## Features:
- Clicking on the different status in the table header filters the submission by status. 
- Can sort based on newest/oldest date submitted.
- Clicking anywhere on the row will take you to the preprint detail page.
- Enable pagination if the number of records exceeds 10.

# Screenshots:
image
image
image


app/components/accepted-record/template.hbs
```diff
@@ -0,0 +1,30 @@
...
+                    <a class="fa fa-check-circle-o fa-lg"> </a>
+                </td>
+                <td>
+                    <div class="media">
+                        <div class="media-body" title={{record.title}}>
+                            <a href="/preprints/{{theme.id}}/{{record.id}}">
+                                <p class="record-title">{{shorten-text record.title screenWidth}}</p>
+                                <p class="record-body">
+                                    submitted on {{moment-format record.dateCreated "MMMM DD, YYYY"}} by
+                                    {{#each record.node.contributors as |contributor|}}
```
chrisseto: @laurenbarker were we going to truncate the list of contributors at some point? RPP would look a bit crazy.

app/components/accepted-record/template.hbs
```diff
@@ -0,0 +1,30 @@
...
+                                        {{contributor.users.fullName}}
+                                        {{#unless (eq contributor record.node.contributors.lastObject)}}
```
chrisseto: A slightly nicer way to do this (IMO) is to add a `,` before the name if the [current index](https://www.emberjs.com/api/ember/2.14/classes/Ember.Templates.helpers/methods/each?anchor=each) is not 0.
It's a touch less code and *slightly* faster.

app/components/moderation-list/component.js
```diff
@@ -0,0 +1,76 @@
...
+ * ```
+ * @class moderation-list
+ **/
+export default Ember.Component.extend({
+    store: Ember.inject.service(),
+    params: Ember.inject.service('pager'),
+    theme: Ember.inject.service(),
+    screenWidth: 0,
+    loading: false,
+    sortOptions: Ember.computed( function() { // Sort options for moderation list
```
chrisseto: Why is this a computed?

app/components/moderation-list/component.js
```diff
@@ -0,0 +1,76 @@
...
+                this.get('params').setAsc();
+                this.send('pageChanged', 1);
+            }
+        },
+        pageChanged: function (current) {
+            this.set('loading', true);
+            const state = this.get('buttonType');
+            this.get('params').setPage(current, state);
+            let requestParams = this.get('params');
+            requestParams["filter[provider]"] = this.get('theme.id');
```
chrisseto: Ember can [bind to query parameters](https://guides.emberjs.com/v2.14.0/routing/query-params/), there's no need to handle this manually.

app/components/moderation-list/component.js
```diff
@@ -0,0 +1,76 @@
...
+            requestParams['filter[reviews_state]'] = state;
+            requestParams['embed'] = 'node';
+            this.get('store').query('preprint', requestParams)
+                .then((results) => {
+                    const records = 'model.'+ state +'Preprints.records';
```
chrisseto: nitpick: When adding a value in the middle of 2 strings, [interpolation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) is much more understandable.

app/components/moderation-list/component.js
chrisseto: If this can't be moved into the controller, you can use the [mut helper](https://www.emberjs.com/api/ember/2.14/classes/Ember.Templates.helpers/methods/mut?anchor=mut) to pass the bindings to the component.

app/components/moderation-list/component.js
```diff
@@ -0,0 +1,76 @@
...
+    actions: {
+        buttonPressed: function(target){
+            Ember.$('.btn-toolbar button[data-target="' + this.get('params').getButton() + '"]').removeClass('activated');
+            this.set('buttonType', target);
+            this.get('params').setButton(target);
+            Ember.$('.table tr').css('display', 'none');
+            Ember.$('.table tr[data-status="' + target + '"]').fadeIn('fast');
+        },
+        sortSelected: function (type) {
+            if (type == 'newest'){
```
chrisseto: This should be handled by handlebars, not jquery. You can use the [inline variant of the if helper](https://www.emberjs.com/api/ember/2.14/classes/Ember.Templates.helpers/methods/mut?anchor=if) to make it a bit easier.

app/components/moderation-list/template.hbs
```diff
@@ -0,0 +1,57 @@
...
+                            {{else if (eq buttonType 'pending')}}
+                                {{pending-record model=model screenWidth=screenWidth}}
+                            {{else}}
+                                {{rejected-record model=model screenWidth=screenWidth}}
+                            {{/if}}
+                        </div>
+                    </div>
+                    {{#if (and (gt model.pendingPreprints.meta.total 1) (eq buttonType 'pending'))}}
+                        <div class="pull-right text-right">
+                            {{pagination-pager count=model.pendingPreprints.meta.total current=params.pendingpage change=(action 'pageChanged')}}
```
chrisseto: It might be a bit nicer to have a computed that returns the appropriate `count` and `current` rather than doing a switch in handlebars.

app/components/pending-record/component.js
```diff
@@ -0,0 +1,17 @@
+import Ember from 'ember';
```
chrisseto: Is there any reason all the `*-record` components couldn't be merged into a single one that accepts a `state` parameter?

app/routes/provider.js
```diff
@@ -0,0 +1,62 @@
+import Ember from 'ember';
+import config from 'ember-get-config';
+
+const providers = config.Reviews.providers.slice(1);
```
chrisseto: These constants are not being used anywhere

app/routes/provider.js
```diff
@@ -0,0 +1,62 @@
...
+                window.location.pathname = pathname.replace(pathRegex, `/preprints/${slugLower}`);
+            }
+        } else {
+            this.replaceWith('page-not-found');
+        }
+    },
+    model(params) {
+        params["filter[provider]"] = this.get('theme.id');
+        delete params["page"];
+        return Ember.RSVP.hash({
```
chrisseto: The model for this route should be just the provider. This logic would belong in the [index route](https://guides.emberjs.com/v2.14.0/routing/defining-your-routes/#toc_index-routes)

app/routes/provider.js
```diff
@@ -0,0 +1,62 @@
...
+                const {pathname} = window.location;
+                const pathRegex = new RegExp(`^/preprints/${slug}`);
+                window.location.pathname = pathname.replace(pathRegex, `/preprints/${slugLower}`);
+            }
+        } else {
+            this.replaceWith('page-not-found');
+        }
+    },
+    model(params) {
+        params["filter[provider]"] = this.get('theme.id');
```
chrisseto: The provider should be pulled from `params.slug` not the theme

app/services/pager.js
```diff
@@ -0,0 +1,27 @@
+import Ember from 'ember';
+
+export default Ember.Service.extend({
```
chrisseto: This looks like logic that would be better encapsulated in a component

app/services/theme.js
```diff
@@ -22,6 +22,14 @@ export default Ember.Service.extend({
     // The id of the current provider
     id: config.Reviews.defaultProvider,
 
+    provider: Ember.computed('id', function() {
```
chrisseto: Is there any reason to have a theme service? Branding is not part of the MVP.

app/styles/app.scss
```diff
@@ -154,6 +154,47 @@
   width: 100%;
 }
 
+
+.btn-filter {
```
chrisseto: If you have styling that is specific to a page, create a separate `.scss` file and import it into this one.

app/styles/app.scss
```diff
@@ -154,6 +154,47 @@
...
+  background-color: Transparent;
+  background-repeat:no-repeat;
+  cursor:pointer;
+  overflow: hidden;
+  outline:none;
+}
+
+.panel-body {
```
chrisseto: You'll want to avoid using generic names if this styling is for a specific element or page.

Using SCSS, you can make it a bit less verbose by using nesting

```scss
.review-list {
    heading {
        ....
    }
}
```

app/styles/app.scss
```diff
@@ -154,6 +154,47 @@
...
+  .btn:hover {
+    text-decoration: none;
+    outline: none;
+  }
+  .btn:focus {
+    font-weight: bold;
+  }
+}
+
+.moderation-list {
```
chrisseto: Let's stick with "reviews" rather than "moderation", for consistency.

app/templates/provider/moderation.hbs
```diff
@@ -0,0 +1,2 @@
+{{moderation-base provider-title='engrXiv' provider-id='engrxiv' unread-count=model.pendingPreprints.meta.pagination.total active='Moderation'}}
```
chrisseto: Hard coded values?

app/templates/provider/settings.hbs
```diff
@@ -0,0 +1 @@
+{{moderation-base provider-title='engrXiv' provider-id='engrxiv' unread-count=10 active='Settings'}}
```
chrisseto: Hard coded values?

config/environment.js
```diff
@@ -34,19 +34,64 @@ module.exports = function(environment) {
...
-        providers: [
-            // OSF must be the first provider
-            {
-                id: 'osf',
-                logoSharing: {
-                    path: '/assets/img/provider_logos/osf-dark.png',
-                    type: 'image/png',
-                    width: 363,
-                    height: 242
+            providers: [
```
chrisseto: Why do we need this list?

app/components/moderation-list/component.js
```diff
@@ -0,0 +1,76 @@
+import Ember from 'ember';
+/**
+ * Provides a list of pending, accepted, and rejected preprints. Provides filtering by preprint state and sorting based
+ * on the preprint creation date. If number of records in each type exceeds 10, pagination is enabled. Current page for
+ * each state is tracked and used by the paginator using the added pager service . Pagination is using the
+ * pagination-pager addon (source: https://github.com/knownasilya/pagination-pager).
```
chrisseto: Could we not use the [osf-paginator component](https://github.com/CenterForOpenScience/ember-osf/tree/develop/addon/components/osf-paginator)?

app/templates/provider/moderation.hbs
chrisseto: You shouldn't need to add this components to these pages/routes. If you place them in `provider.hbs` with an `{{outlet}}` ember will render all subpages/subroutes in between them.

app/breakpoints.js
```diff
@@ -0,0 +1,5 @@
+export default {
+  mobile:  '(max-width: 767px)',
```
chrisseto: [Bootstrap defines media break points](https://github.com/twbs/bootstrap-sass/blob/master/templates/project/_bootstrap-variables.sass#L287), we should use those if required.

app/components/accepted-record/template.hbs
```diff
@@ -0,0 +1,30 @@
...
+        {{#each model.acceptedPreprints.records as |record|}}
+            <tr data-status={{record.reviewsState}}>
+                <td class="state">
+                    <a class="fa fa-check-circle-o fa-lg"> </a>
+                </td>
+                <td>
+                    <div class="media">
+                        <div class="media-body" title={{record.title}}>
+                            <a href="/preprints/{{theme.id}}/{{record.id}}">
+                                <p class="record-title">{{shorten-text record.title screenWidth}}</p>
```
chrisseto: Have we come to a consensus on word wrapping vs truncation?

app/components/accepted-record/template.hbs
```diff
@@ -0,0 +1,30 @@
...
+                                <p class="record-body">
+                                    submitted on {{moment-format record.dateCreated "MMMM DD, YYYY"}} by
```
chrisseto: We should be consistent with how the rest of the OSF displays datetimes. It looks closer to `YYYY-MM-DD H:M` (2017-01-13 01:18 PM). It's probably specified in ember OSF somewhere

chrisseto (changes requested):

app/components/accepted-record/template.hbs
laurenbarker: Yes, I don't think I included that in the mockups (my bad). Truncated after first 3 -- should do the same thing as the share interface where we put "+10" or whatever after they get truncated.

app/services/theme.js
aaxelb: Parts of ember-osf will break without `theme.provider` and `theme.isProvider`. We need to have it (because that's less work than fixing ember-osf), but probably shouldn't use it for now.

app/components/accepted-record/template.hbs
laurenbarker: How the date is displayed on the preprints detail page:
`{{t logDateLabel}}: {{moment-format logDate "MMMM DD, YYYY"}} | {{t "content.header.last_edited"}}: {{moment-format model.dateModified "MMMM DD, YYYY"}}`

app/components/accepted-record/template.hbs
laurenbarker: Truncation for now but we can take another look once everything is finished to see if that is the appropriate solution for all screen sizes. Might want to create a ticket for the next sprint so we don't forget.

chrisseto:
Is there an overlap between this PR and https://github.com/CenterForOpenScience/ember-osf-reviews/pull/2?

app/components/pending-record/component.js
sheriefvt: currently, they are almost same, but in the future, we will have different wording, requirements for example accepted and rejected will require info from the reviews log and reviewer name

app/components/moderation-list/component.js
sheriefvt: Currently, pagination-pager is being used by registries and preprints. [osf-paginator](https://github.com/CenterForOpenScience/ember-osf/tree/develop/addon/components/osf-paginator ) is not used.

app/components/moderation-list/component.js
sheriefvt: Forget to add a Todo, but planning to pass i18n.locale to it when translation is ready

app/components/moderation-list/component.js
chrisseto: 👍

app/components/accepted-record/template.hbs
```diff
@@ -0,0 +1,33 @@
...
+                    <a class="fa fa-check-circle-o fa-lg"> </a>
+                </td>
+                <td>
+                    <div class="media">
+                        <div class="media-body" title={{record.title}}>
+                            <a href="/preprints/{{theme.id}}/{{record.id}}">
+                                <p class="record-title">{{shorten-text record.title screenWidth}}</p>
+                                <p class="record-body">
+                                    {{t 'moderation_list.record.submitted_on'}} {{moment-format record.dateCreated "MMMM DD, YYYY"}} {{t 'moderation_list.record.by'}}
+                                    {{#each record.node.contributors as |contributor index| ~}}
```
chrisseto: Rather than `....`, the mock ups display `+ X` where X is however many contributors are not displayed. Sounds like a good case for a component.

app/components/pending-record/component.js
chrisseto: I think it would be cleaner/more maintainable to encapsulate all of that logic into a single component.

app/components/pending-record/template.hbs
```diff
@@ -0,0 +1,32 @@
...
+                    <a class="fa fa-hourglass-o fa-lg"> </a>
+                </td>
+                <td>
+                    <div class="media">
+                        <div class="media-body" title={{record.title}}>
+                            <a href="/preprints/{{theme.id}}/{{record.id}}">
+                                <p class="record-title">{{shorten-text record.title screenWidth}}</p>
+                                <p class="record-body">
+                                    {{t 'moderation_list.record.submitted_on'}} {{moment-format record.dateCreated "MMMM DD, YYYY"}} {{t 'moderation_list.record.by'}}
+                                    {{#each record.node.contributors as |contributor index| ~}}
```
chrisseto: This should be replaced by the previously mentioned component.

app/components/rejected-record/template.hbs
```diff
@@ -0,0 +1,32 @@
...
+                    <a class="fa fa-times-circle-o fa-lg"> </a>
+                </td>
+                <td>
+                    <div class="media">
+                        <div class="media-body" title={{record.title}}>
+                            <a href="/preprints/{{theme.id}}/{{record.id}}">
+                                <p class="record-title">{{shorten-text record.title screenWidth}}</p>
+                                <p class="record-body">
+                                    {{t 'moderation_list.record.submitted_on'}} {{moment-format record.dateCreated "MMMM DD, YYYY"}} {{t 'moderation_list.record.by'}}
+                                    {{#each record.node.contributors as |contributor index| ~}}
```
chrisseto: This should be replaced by the previously mentioned component.

app/routes/provider.js
```diff
@@ -0,0 +1,47 @@
...
+export default Ember.Route.extend(queryParamsMixin, {
+    theme: Ember.inject.service(),
+    active: 'Moderation',
+    model(params, transition) {
+        delete params['page'];
+        const {slug = ''} = transition.params.provider;
+        const slugLower = slug.toLowerCase();
+        this.set('theme.id', slugLower);
+        return Ember.RSVP.hash({
+            acceptedPreprints: this.store.query('preprint',
```
chrisseto: This is fine for the moment. @aaxelb is there anyway we could include all the counts in a single request?

app/styles/app.scss
```diff
@@ -154,6 +154,47 @@
...
+  overflow: hidden;
+  outline:none;
+}
+
+.panel-body {
+  padding: 0;
+  box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
+}
+
+.heading {
```
chrisseto: Please use a more specific name.

config/deprecation-workflow.js
```diff
@@ -0,0 +1,7 @@
+window.deprecationWorkflow = window.deprecationWorkflow || {};
```
chrisseto: I think we should leave these warnings, eventually they will annoy someone enough that they'll get fixed.

app/styles/app.scss
chrisseto: Re: ^

app/styles/app.scss
```diff
@@ -210,3 +251,86 @@ li.active > a {
 .column-offset {
   height: 120px;
 }
+
+.state {
```
chrisseto: These should have more specific names and use SCSS variables.

app/styles/app.scss
```diff
@@ -210,3 +251,86 @@ li.active > a {
...
+  }
+  a:hover {
+    text-decoration: none;
+  }
+  a:active {
+    text-decoration: none;
+  }
+}
+
+@media screen and (max-width: 480px) {
```
chrisseto: Use Bootstraps media query variables here

app/components/accepted-record/template.hbs
```diff
@@ -0,0 +1,33 @@
...
+    <tbody>
+        {{#each model.acceptedPreprints.records as |record|}}
+            <tr data-status={{record.reviewsState}}>
+                <td class="state">
+                    <a class="fa fa-check-circle-o fa-lg"> </a>
+                </td>
+                <td>
+                    <div class="media">
+                        <div class="media-body" title={{record.title}}>
+                            <a href="/preprints/{{theme.id}}/{{record.id}}">
```
chrisseto: Is there another way to get the provider ID? `{{theme.id}}` seems a bit hacky.

app/components/moderation-list/template.hbs
```diff
@@ -0,0 +1,57 @@
...
+                            {{else if (eq buttonType 'pending')}}
+                                {{pending-record model=model screenWidth=screenWidth}}
+                            {{else}}
+                                {{rejected-record model=model screenWidth=screenWidth}}
+                            {{/if}}
+                        </div>
+                    </div>
+                    {{#if (and (gt model.pendingPreprints.meta.total 1) (eq buttonType 'pending'))}}
+                        <div class="pull-right text-right">
+                            {{pagination-pager count=model.pendingPreprints.meta.total current=pendingpage change=(action 'pageChanged')}}
```
chrisseto: I think using a computed property here instead of switching in the template would be nicer

app/components/moderation-list/template.hbs
```diff
@@ -0,0 +1,57 @@
...
+                                <span>
+                                    <button class="btn" onclick={{action "sortSelected" '-date_created'}}><a class='{{if (eq sort '-date_created') 'fa fa-check'}}'></a> {{t 'moderation_list.newest'}}</button>
+                                    <button class="btn" onclick={{action "sortSelected" 'date_created'}}><a class='{{if (eq sort 'date_created') 'fa fa-check'}}'></a> {{t 'moderation_list.oldest'}}</button>
+                                </span>
+                            </ul>
+                        </div>
+                    </div>
+                </div>
+                {{#if loading}}
+                    <div class="text-center p-v-md" aria-label='Loading data'>
```
chrisseto: HTML attributes should be in double quotes, not single. Mostly just for convention but it also makes it a bit easier to distinguish between static and bound attributes.

app/components/moderation-list/component.js
```diff
@@ -0,0 +1,68 @@
...
+        },
+        pageChanged: function (current) {
+            this.set('loading', true);
+            const state = this.get('buttonType');
+            this.set('page', current);
+            this.set('filter', {'provider': this.get('theme.id'), 'reviews_state': state});
+            this.set(state +'page', current);
+            this.set('sort', this.get('sortType'));
+            this.get('store').query('preprint', this.get('params'))
+                .then((results) => {
```
chrisseto: This should be at the same level of indentation as the rest of these lines.

chrisseto (changes requested):

sheriefvt:
@chrisseto thanks for ur feedback..done with changes

chrisseto:
@sheriefvt can you fix the merge conflicts?

sheriefvt:
@chrisseto fixed the merge conflict

app/components/preprint-record/component.js
```diff
@@ -0,0 +1,27 @@
...
+ *    screenWidth=screenWidth
+ * }}
+ * ```
+ * @class accepted-record
+ **/
+export default Ember.Component.extend({
+    theme: Ember.inject.service(),
+    preprintRecords: Ember.computed('data', 'type', function () {
+        const type = this.get('type');
+        return type == 'accepted' ? this.get('data.acceptedPreprints.records') : type == 'pending' ?
```
chrisseto: It would be better to have an object defined with all the possible values and then access the correct key.
We should be using constant values for these variables as well. I think we could just drop them in the config file.
```javascript
{
 key1: value1,
 key2: value2,
}[keyVar];
```

sheriefvt:
@chrisseto changes done

app/controllers/provider/moderation.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
+
+export default Ember.Controller.extend({
+    acceptedpage: 1,
```
chrisseto: What are the purposes of these values?

app/routes/provider.js
```diff
@@ -8,15 +9,39 @@ import Ember from 'ember';
 /**
  * @class Provider Route Handler
  */
-export default Ember.Route.extend({
+export default Ember.Route.extend(queryParamsMixin, {
```
chrisseto: This would be better suited as the `index` route, the counts will only be seen on this page, correct?

config/deprecation-workflow.js
chrisseto: ping

tests/helpers/responsive.js
```diff
@@ -0,0 +1,60 @@
+import Ember from 'ember';
```
chrisseto: Looks like this file can be removed?

chrisseto (changes requested):

app/controllers/provider/moderation.js
sheriefvt: track the current page for each list type (pending, accepted or rejected)

config/deprecation-workflow.js
sheriefvt: I removed it from the list but kept the file in config for future use. The warnings are showing now

tests/helpers/responsive.js
sheriefvt: ok

app/routes/provider.js
sheriefvt: we need it for this page only

sheriefvt:
@chrisseto pushed latest changes..thanks

PR #27, opened 2017-09-19
[Fix] Fix scroll when change routes

sheriefvt:


laurenbarker (dismissed):
Looks good, thanks!

laurenbarker:
Will merge once tests pass 👍

sheriefvt:
@laruen sorry i just made tiny change..remove comment from test

laurenbarker (approved):
👍

jamescdavis:
I know this is already merged, and this is totally a drive by, but I'd like to throw out the idea of resetting scroll in the didTransition() hook of the router rather than using a mixin or a base route. It's one of the refactors I'd like to do for preprints. Try it out with [this twiddle](https://ember-twiddle.com/a841e05d3222e60b4bcd9b681dd4e09d?openFiles=router.js).

laurenbarker:
@jamescdavis is there an issue with adding the reset to the didTransition() hook in the base route? Or issue with using a base route in general? It just seems nicer than remembering to add the reset to every new route.

laurenbarker:
Oh, I read this wrong. Router would work well 👍

PR #28, opened 2017-09-20
[MOD-86][Fix] Fix scroll when transition- followup

sheriefvt:
# Ticket:
https://openscience.atlassian.net/browse/MOD-65 
Follow-up PR for another PR #27 
## What:
Reset scroll on page transition.
## Why:
A user would expect to start at the top of pages and not the middle/end.


laurenbarker:
@sheriefvt, would you mind rebasing your branch on develop?

sheriefvt:
@laurenbarker I created another PR #31 instead of this one

PR #31, opened 2017-09-20
[MOD-86][Fix] Fix scroll reset on route transition

sheriefvt:


app/router.js
```diff
@@ -3,7 +3,19 @@ import config from './config/environment';
 
 const Router = Ember.Router.extend({
     location: config.locationType,
-    rootURL: config.rootURL
+    rootURL: config.rootURL,
+    willTransition(oldInfos, newInfos) {
+        const disableList = ['settings', 'moderation'];
```
aaxelb: This should be an all-caps constant at the top level of the file, with a more descriptive name (maybe `SCROLL_RESET_DISABLED_ROUTES`? or something more concise)

app/router.js
```diff
@@ -3,7 +3,19 @@ import config from './config/environment';
...
+        const prevRoute = oldInfos.length ? oldInfos[oldInfos.length - 1].name.split('.') : null;
```
aaxelb: It'd be nicer to use `Ember.Router._routePath(infos)` (which I discovered looking at the [parent `willTransition`](https://github.com/emberjs/ember.js/blob/v2.14.1/packages/ember-routing/lib/system/router.js#L331) -- this feels like a case where using a private method is better than doing it manually), and have the list be `['preprints.provider.moderation', 'preprint.provider.settings']`

app/router.js
```diff
@@ -3,7 +3,19 @@ import config from './config/environment';
...
+        const prevName = prevRoute != null ? prevRoute[prevRoute.length - 1] : null;
+        const currentRoute = newInfos.length ? newInfos[newInfos.length - 1].name.split('.') : null;
+        const currentName = currentRoute != null ? currentRoute[currentRoute.length - 1] : null;
+        const disableResetScroll = disableList.includes(prevName) && disableList.includes(currentName);
+        if (!disableResetScroll) {
+            this._super(...arguments);
```
aaxelb: I think it should always call `_super`, even if it doesn't scroll to the top. Unless there's something I'm missing?

aaxelb (changes requested):
Looks good! Just a few cleanup suggestions.

sheriefvt:
@aaxelb thanks for ur feedback, done with changes

app/router.js
```diff
@@ -1,9 +1,21 @@
...
+const SCROLL_RESET_DISABLED_ROUTES = ['preprints.provider.moderation', 'preprints.provider.settings'];
+
 const Router = Ember.Router.extend({
     location: config.locationType,
-    rootURL: config.rootURL
+    rootURL: config.rootURL,
+    willTransition(oldInfos, newInfos) {
+        this._super(...arguments);
+        const prevPath = Ember.Router._routePath(oldInfos);  //Previous route
+        const newPath = Ember.Router._routePath(newInfos);   //New route
```
aaxelb: You can remove these two comments, I think it's pretty clear what these are. If you want, could rename the variables to `oldRoute` and `newRoute` or something like that.

aaxelb (dismissed):
Looks good! 👍

sheriefvt:
@aaxelb done

aaxelb (approved):

PR #41, opened 2017-09-26
[MOD-105][Fix] Hide non-set-up providers on moderators' dashboard

aaxelb:


laurenbarker (approved):
Tested, works as expected 💯

PR #42, opened 2017-09-28
[MOD-66][Feature] Use Google Analytics

sheriefvt:
## What:
Add analytics to the reviews app.
## Changes

- Setup analytics.
- Add tracking to page/route views.
- Add tracking to the pagination on the moderation list page.
- Update tests.

## Ticket 
https://openscience.atlassian.net/browse/MOD-66


laurenbarker (dismissed):
Tested locally; looks good! 
@binoculars just to confirm, we don't need to do anything special here right? You guys add the GA token when deploying?

config/environment.js
```diff
@@ -31,6 +31,22 @@ module.exports = function(environment) {
         i18n: {
             defaultLocale: 'en'
         },
+        metricsAdapters: [
+            {
+                name: 'GoogleAnalytics',
+                environments: ['production'], // Add development to try/inspect in development.
+                config: {
+                    // Remove id before committing.
+                    id: '',
```
binoculars: Still need this to be `process.env.GOOGLE_ANALYTICS_ID` for now

binoculars (changes requested):
We will replace all of the ember configs when we go to prod, but for now, we still need to keep the environment variables in the configuration. See https://github.com/CenterForOpenScience/ember-preprints/blob/develop/config/environment.js (minus keen)

app/router.js
```diff
@@ -29,7 +29,22 @@ const Router = Ember.Router.extend({
         if (!this.get('disableResetScroll')) {
             window.scrollTo(0, 0);
         }
+    },
+
+    didTransition(){
```
aaxelb: Combine this with the `didTransition` above (thanks Travis!)

app/controllers/preprints/provider/moderation.js
```diff
@@ -13,6 +14,12 @@ export default Ember.Controller.extend({
         },
         pageChanged(page) {
             this.set('page', page);
+            Ember.get(this, 'metrics')
+                .trackEvent({
+                    category: 'button',
+                    action: 'click',
+                    label: `Request page ` + this.get('page') + ` of ` + this.get('status') + ` list`
```
aaxelb: I think it'd be good to sort of namespace labels by page/component, like ember-preprints does: "Moderation List - Request page..."

aaxelb (changes requested):

app/router.js
sheriefvt: @aaxelb I think this happened when merged with develop

app/router.js
```diff
@@ -7,7 +7,7 @@ const SCROLL_RESET_DISABLED_ROUTES = ['preprints.provider.moderation', 'preprint
 const Router = Ember.Router.extend({
     location: config.locationType,
     rootURL: config.rootURL,
-
+    metrics: Ember.inject.service(),
```
aaxelb: Minor: Put injections at the top of the class

aaxelb (dismissed):
Looks good to me (with one nitpick), if there are no other devopsy needs @binoculars

laurenbarker:
@sheriefvt, when you get a chance could you fix the merge conflicts?

sheriefvt:
@laurenbarker yes sure

PR #43, opened 2017-09-29
[MOD-113][Fix] Fix date on moderation list page

laurenbarker:
## Purpose
All dates on the moderation list page were the current date due to a rename mistake.

## Changes
* less than a day show relative time
* update wording on accepted and rejected page to be more relevant

screen shot 2017-09-29 at 11 42 05 am
screen shot 2017-09-29 at 11 42 23 am

### DIFF
--- a/app/components/moderation-list-row/component.js
+++ b/app/components/moderation-list-row/component.js
@@ -1,4 +1,25 @@
 import Ember from 'ember';
+import moment from 'moment';
+
+const PENDING = 'pending';
+const ACCEPTED = 'accepted';
+const REJECTED = 'rejected';
+
+const ACTION_LABELS = Object.freeze({
+    [PENDING]: {
+        gtDay: 'components.moderation-list-row.submission.submitted_on',
+        ltDay: 'components.moderation-list-row.submission.submitted',
+    },
+    [ACCEPTED]: {
+        gtDay: 'components.moderation-list-row.submission.was_accepted_on',
+        ltDay: 'components.moderation-list-row.submission.was_accepted',
+    },
+    [REJECTED]: {
+        gtDay: 'components.moderation-list-row.submission.was_rejected_on',
+        ltDay: 'components.moderation-list-row.submission.was_rejected',
+    }
+});
+
 
 export default Ember.Component.extend({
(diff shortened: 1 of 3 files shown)


app/components/moderation-list-row/component.js
```diff
@@ -11,7 +32,25 @@ export default Ember.Component.extend({
...
     },
 
+    gtDay: Ember.computed('submission.dateLastTransitioned', function() {
+        return moment().diff(this.get('submission.dateLastTransitioned'), 'days') > 1;
+    }),
+
+    relevantDate: Ember.computed('submission.dateLastTransitioned', 'gtDay', function() {
+        return this.get('gtDay') ?
+            moment(this.get('submission.dateLastTransitioned')).format('MMMM DD, YYYY') :
+            moment(this.get('submission.dateLastTransitioned')).fromNow();
```
sheriefvt, line 42: I like this change, but concerned on how the senetece generated by fromNow() (e.g. 44 seconds ago) will be translated.

sheriefvt (changes requested):
Looks good 👍 , only one concern/change. Feel free to ignore 🐼  if you want

app/components/moderation-list-row/component.js
sheriefvt: also, do we need to check with product on wording changes

laurenbarker: Oh good point. I hadn't considered that.

laurenbarker: Okay, [moment supports internationalization](https://momentjs.com/docs/#/i18n/) so if there are multiple languages I think we would just need to set the language in https://github.com/CenterForOpenScience/ember-osf-reviews/blob/develop/app/routes/application.js.

laurenbarker: I'll double check on the wording changes. We couldn't use what was in the mockups due to the difficulty of getting the moderator's name.

laurenbarker: Wording approved by product and MOD-115 added to backlog to update the wording with the moderator's name.

PR #49, opened 2017-10-02
[MOD-107][Improvement] Show not set up yet error

laurenbarker:
## Changes
screen shot 2017-10-03 at 3 07 03 pm


app/routes/index.js
```diff
@@ -9,6 +9,18 @@ export default Ember.Route.extend({
         page: { refreshModel: true }
     },
 
+    beforeModel() {
```
aaxelb: This same `beforeModel` is in `preprint.provider`, so it's run for every route in our app. Could we put it in the application route instead of having it twice?

laurenbarker: I'm going to refactor some stuff then come back to this.

aaxelb (approved):

PR #51, opened 2017-10-03
[MOD-103][Fix] Check provider permissions

laurenbarker:
Add additional permissions check to catch moderators trying to access a provider of which they do not have permissions.

### DIFF
--- a/app/routes/preprints/provider.js
+++ b/app/routes/preprints/provider.js
@@ -35,7 +35,7 @@ export default Ember.Route.extend({
         if (!model.get('permissions').contains('view_submissions')) {
             this.replaceWith('forbidden');
         } else if (!model.get('reviewsWorkflow') && transition.targetName !== 'preprints.provider.setup') {
-            return this.replaceWith('preprints.provider.setup', model);
+            this.replaceWith('preprints.provider.setup', model);
         }
     },
 });
--- a/app/routes/preprints/provider.js
+++ b/app/routes/preprints/provider.js
@@ -32,6 +32,10 @@ export default Ember.Route.extend({
     },
 
     afterModel(model, transition) {
-        if (!model.get('reviewsWorkflow') && transition.targetName !== 'preprints.provider.setup') return this.replaceWith('preprints.provider.setup', model);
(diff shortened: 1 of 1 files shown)


app/routes/preprints/provider.js
```diff
@@ -32,6 +32,10 @@ export default Ember.Route.extend({
     },
 
     afterModel(model, transition) {
-        if (!model.get('reviewsWorkflow') && transition.targetName !== 'preprints.provider.setup') return this.replaceWith('preprints.provider.setup', model);
+        if (!model.get('permissions').contains('view_submissions')) {
+            this.replaceWith('forbidden');
+        } else if (!model.get('reviewsWorkflow') && transition.targetName !== 'preprints.provider.setup') {
+            return this.replaceWith('preprints.provider.setup', model);
```
aaxelb: `replaceWith` returns the transition object... does returning that from `afterModel` do anything?

laurenbarker: Yup, `afterModel` is called with a transition object. See [Transitions docs](https://guides.emberjs.com/v2.15.0/routing/preventing-and-retrying-transitions/#toc_aborting-transitions-within-code-model-code-code-beforemodel-code-code-aftermodel-code) or [Redirection docs](https://guides.emberjs.com/v2.15.0/routing/redirection/).

aaxelb: I'm wondering why there's a `return` here, but not two lines up. It [looks like](https://www.emberjs.com/api/ember/2.15/classes/Ember.Route/methods/afterModel?anchor=afterModel) values returned from `afterModel` are ignored unless they're promises.

laurenbarker: Ohhh, I get what you're saying now. I didn't add that second line but it does work. I guess it be useful for the transition to wait for it to finish.

aaxelb: I'm pretty sure returning does nothing, and I feel like it's misleading to include. Would you mind removing it?

aaxelb (dismissed):

laurenbarker:
Removed and tested 👍

PR #53, opened 2017-10-04
[MOD-119][MOD-103][MOD-101] Refactor permission checks

laurenbarker:
## Changes
* [MOD-119] Remove `not-authenticated` route and display not authenticated page on index
    * Checks `session.isAuthenticated` in `application.js`
* [MOD-103][MOD-101] Check `canViewReviews` in `application.js`
    * Keep `view_submissions` check in `provider.js`
    * Keep `set_up_moderation` check in `setup.js`


app/routes/application.js
```diff
@@ -32,5 +35,16 @@ export default Ember.Route.extend(OSFAgnosticAuthRouteMixin, {
         if (locale) {
             this.set('i18n.locale', locale);
         }
+
+        // Check permissions
+        if (!this.get('session.isAuthenticated')) {
+            this.replaceWith('index');
+        } else {
+            this.get('currentUser.user').then((user) => {
```
aaxelb: Ooh, I think *this* is when we want to return something from a model hook. The transition should pause until this promise resolves.

app/routes/application.js
```diff
@@ -32,5 +35,16 @@ export default Ember.Route.extend(OSFAgnosticAuthRouteMixin, {
         if (locale) {
             this.set('i18n.locale', locale);
         }
+
+        // Check permissions
```
aaxelb, line 39: This seems like it'd make more sense in `beforeModel`, so it won't try fetching all the models before redirecting to `index`.

aaxelb (changes requested):
Nice 🍖 

Just a couple small things

app/routes/application.js
laurenbarker: I had it in `beforeModel` originally but session wasn't defined yet so it always said the user wasn't logged in.

app/routes/application.js
laurenbarker: 👍  will try that

app/routes/application.js
aaxelb: Aw, that's annoying

laurenbarker:
@aaxelb, addressed 👗

aaxelb (approved):

PR #54, opened 2017-10-04
[Fix][MOD-120] Update wording generated/returned by moment.js

sheriefvt:
## Problem 
On transition after a preprint submission, preprints would be listed as modified "a few seconds ago", however sometimes it shows in "a few seconds" or "a minute ago". It seems that there is a time difference between client/server.

## Solution
Calculate the absolute time difference within a slack period of 70000 milliseconds to accommodate the time difference between client/server.

Alt Text

🐼 
## Ticket
https://openscience.atlassian.net/browse/MOD-120

### DIFF
--- a/app/components/moderation-list-row/component.js
+++ b/app/components/moderation-list-row/component.js
@@ -44,10 +44,20 @@ export default Ember.Component.extend({
         return moment().diff(this.get('submission.dateLastTransitioned'), 'days') > 1;
     }),
 
+    fewSeconds: Ember.computed('submission.dateLastTransitioned', function() {
+        const timeDiff = Math.abs(moment().diff(this.get('submission.dateLastTransitioned')));
+        if (timeDiff < 70000) { // 70000 milliseconds
+            return 'a few seconds ago';
+        } else {
+            return moment(this.get('submission.dateLastTransitioned')).fromNow();
+        }
+    }),
+
+
     relevantDate: Ember.computed('submission.dateLastTransitioned', 'gtDay', function() {
         return this.get('gtDay') ?
             moment(this.get('submission.dateLastTransitioned')).format('MMMM DD, YYYY') :
-            moment(this.get('submission.dateLastTransitioned')).fromNow();
+            this.get('fewSeconds');
     }),
 
     //translations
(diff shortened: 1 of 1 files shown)


app/components/moderation-list-row/component.js
```diff
@@ -44,10 +44,20 @@ export default Ember.Component.extend({
         return moment().diff(this.get('submission.dateLastTransitioned'), 'days') > 1;
     }),
 
+    fewSeconds: Ember.computed('submission.dateLastTransitioned', function() {
+        const timeDiff = Math.abs(moment().diff(this.get('submission.dateLastTransitioned')));
+        if (timeDiff < 70000) { // 70000 milliseconds
```
aaxelb: Is there a reason for the magic 70000?

aaxelb (changes requested):
The problem seems to be when `dateLastTransitioned` is slightly in the future -- I think it would be simpler to do something like `min(datelastTransitioned, now)`, replacing any future values with now. Does that seem reasonable?

app/components/moderation-list-row/component.js
sheriefvt: I used absolute value so it can handle past or future differences. Sometimes datelasttransitioned seems slightly in past too.

aaxelb: Slightly in the past is fine -- what we'd expect, actually. But if it's somehow more than 70 seconds in the future, it might display like "a minute from now", which is bad.

sheriefvt: 7000 seems to cover the possible differences after several trials, which slight more than a minute

sheriefvt: so when I tried to change preprint status to accepted or rejected.. was showing a minute ago... while the transition already just happened immediately

aaxelb: I think "a minute ago" is fine, it's close enough to true. The bug is when it displays "in a few seconds", or anything that implies the transition is in the future.

sheriefvt: ok 👍

aaxelb (approved):

PR #56, opened 2017-10-05
[MOD-125][Fix] Force reload moderation list query

aaxelb:
Uses a [slightly modified](https://github.com/cos-forks/ember-data-has-many-query/commit/cc586f54606d167dc53e3f85d131231d572d9745) version of the `ember-data-has-many-query` addon which allows manually forcing a reload.


laurenbarker (dismissed):
Works 💯 . Should maybe put a comment?

PR #58, opened 2017-10-06
[MOD-84][MOD-126] Empty action lists

aaxelb:
Handle empty action lists, and do a little cleanup. I ended up fixing a few things I thought I broke but in fact were already broken.

- upgrade to Ember 2.15
- add `loading-indicator` component, use it instead of all those silly divs
- don't use ember-data-has-many-query to query preprints on the moderation list page
    - it breaks when the first query returns 0 results
    - instead I added a simpler but largely equivalent `query` function that sends a request manually and pushes the results into the store
    - also fixes MOD-126!

screen shot 2017-10-06 at 4 27 40 pm
screen shot 2017-10-06 at 4 32 37 pm

https://openscience.atlassian.net/browse/MOD-84
https://openscience.atlassian.net/browse/MOD-126


app/components/loading-indicator/template.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="ball-pulse ball-dark {{classes}}" aria-label="Loading data">
```
laurenbarker: `ball-dark` should probably be a passed in class because we don't always want the dark version (e.g. `contributor-list`).

app/components/preprint-status-banner/template.hbs
```diff
@@ -10,7 +10,15 @@
             </span>
         </div>
         <div class="recent-activity">
-            <a href={{creatorProfile}}>{{creatorName}}</a> {{t recentActivityLanguage}} {{moment-format submission.dateLastTransitioned "MMMM DD, YYYY"}}
+            {{#if loadingActions}}
+                {{loading-indicator classes='text-center m-b-lg p-v-md'}}
```
laurenbarker: The extra padding isn't needed when loading:
screen shot 2017-10-09 at 11 12 03 am
screen shot 2017-10-09 at 11 12 38 am

laurenbarker (changes requested):
Just a few tiny things 💯

aaxelb:
@laurenbarker Ready for another look!

laurenbarker:
Clicking on a row before it's fully loaded results in contributors never being loaded:
no-contributors

laurenbarker:
Also weird things are happening when transitioning from the dashboard (at least for me):
odd-transitions

app/routes/preprints/provider/moderation.js
```diff
@@ -4,6 +4,38 @@ import Ember from 'ember';
...
+            }).then(payload => {
+                store.pushPayload(payload);
+                const records = payload.data.map(datum => store.peekRecord(datum.type, datum.id));
+                records.meta = payload.meta;
+                records.links = payload.links;
+                resolve(records);
+            }, reject);
+        } else {
+            const message = `Could not find a link for '${propertyName}' relationship`;
+            console.error(message, 'on model:\n', model);
```
laurenbarker: Travis doesn't like this

laurenbarker (changes requested):
Seems like some weird behavior was introduced with the transition aborted fix.

aaxelb:
@laurenbarker I was able to replicate that first bug (clicking on a row while it's still loading causing contributors not to load) on staging3, seems worth a separate ticket?

The weird transitions to/from the dashboard are because I added `templates/loading.hbs`. I removed it because it does behave weirdly enough that it's better to just be slow. We should maybe think about making more route-specific loading templates, if the model hooks are slow enough on staging/prod.

laurenbarker (approved):

laurenbarker:
Ah, okay. Yeah I'll add a ticket for route specific loading pages and the contributors bug.

PR #60, opened 2017-10-10
[Fix][MOD-137] Contributors not showing when click row before loading

sheriefvt:
## Problem
Contributors won't show if you click the moderation list row before it loads

## Solution

- Disable the link-to tell the needed data is loaded. 
- Adjust the loading-indicator behavior 

## Ticket
https://openscience.atlassian.net/browse/MOD-137


aaxelb (approved):
This avoids the bug just fine, so it's good for now, but I could see users being slightly annoyed they can't click through if loading contributors goes slowly. I may try looking into fixing the underlying bug if/when I have a moment.

laurenbarker:
@aaxelb https://openscience.atlassian.net/browse/MOD-140

PR #61, opened 2017-10-10
[Fix][MOD-139] Paginations controls not showing

sheriefvt:
## Problem
Pagination controls are not showing when there are more than 10 submissions.

## Solution
Response meta total was undefined. Replace with adjusted counts in reviews_state_counts

## Ticket
https://openscience.atlassian.net/browse/MOD-139

### DIFF
--- a/app/routes/preprints/provider/moderation.js
+++ b/app/routes/preprints/provider/moderation.js
@@ -56,7 +56,7 @@ export default Ember.Route.extend({
         }).then(response => {
             return {
                 submissions: response.toArray(),
-                totalPages: Math.ceil(response.meta.reviews_state_counts[params.status] / 10),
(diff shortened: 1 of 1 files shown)


app/routes/preprints/provider/moderation.js
```diff
@@ -56,7 +56,7 @@ export default Ember.Route.extend({
         }).then(response => {
             return {
                 submissions: response.toArray(),
-                totalPages: response.meta.total,
+                totalPages: Math.ceil(response.meta.reviews_state_counts[params.status] / 10),
```
aaxelb: The pagination meta is available under `response.links.meta`, so you could do `Math.ceil(response.links.meta.total / response.links.meta.per_page)` instead

aaxelb (changes requested):

aaxelb (approved):

PR #64, opened 2017-10-11
[Feature] Add queryHasMany method to the store

aaxelb:
Use `store.queryHasMany(model, 'propertyName', { param: value })` instead of `model.query('propertyName', { param: value })`


sheriefvt (approved):
looks good to me

PR #68, opened 2017-10-13
[Fix][MOD-141] Fix see more link dashboard

sheriefvt:
## Problem
The "See more" link is not displayed on the dashboard, even when there are more actions available. Make it display whenever there are more actions.

## Ticket
https://openscience.atlassian.net/browse/MOD-141

### DIFF
--- a/app/components/action-feed/component.js
+++ b/app/components/action-feed/component.js
@@ -19,6 +19,8 @@ export default Component.extend({
     currentUser: service(),
     classNames: ['action-feed'],
     page: 1,
+    listLoading: false,
+
     errorMessage: t('components.action-feed.error_loading'),
 
     moreActions: computed('totalPages', 'page', function() {
@@ -29,16 +31,18 @@ export default Component.extend({
         this._super(...arguments);
         this.actionsList = [];
         this.loadPage();
+        this.set('listLoading', true);
+        this.actionsList = [];
     },
 
     loadPage() {
         const page = this.get('page');
         this.set('loadingPage', page);
-        this.get('user').query('actions', { page: this.get('page') }).then(
+        this.get('store').queryHasMany(this.get('user'), 'actions', { page: this.get('page') }).then(
             response => this._setPageProperties(response)
             , () => {
                 // Error
-                this.set('loadingPage', null);
+                this.set('listLoading', false);
                 this.get('toast').error(this.get('errorMessage'));
             },
         );
@@ -48,8 +52,8 @@ export default Component.extend({
         if (this.get('loadingPage') === this.get('page')) {
             this.setProperties({
                 actionsList: this.get('actionsList').concat(response.toArray()),
-                totalPages: response.get('meta.total'),
-                loadingPage: null,
+                totalPages: Math.ceil(response.get('links.meta.total') / response.get('links.meta.per_page')),
+                listLoading: false,
             });
         }
     },
--- a/app/components/action-feed/component.js
+++ b/app/components/action-feed/component.js
@@ -19,7 +19,6 @@ export default Component.extend({
     currentUser: service(),
     classNames: ['action-feed'],
     page: 1,
-    listLoading: false,
 
     errorMessage: t('components.action-feed.error_loading'),
 
@@ -29,9 +28,9 @@ export default Component.extend({
 
     init() {
         this._super(...arguments);
-        this.loadPage();
         this.set('listLoading', true);
         this.set('actionsList', []);
+        this.loadPage();
     },
 
     loadPage() {
--- a/app/components/action-feed/component.js
+++ b/app/components/action-feed/component.js
@@ -29,7 +29,6 @@ export default Component.extend({
 
     init() {
         this._super(...arguments);
-        this.actionsList = [];
         this.loadPage();
         this.set('listLoading', true);
         this.actionsList = [];
@@ -37,29 +36,33 @@ export default Component.extend({
 
     loadPage() {
         const page = this.get('page');
-        this.set('loadingPage', page);
-        this.get('store').queryHasMany(this.get('user'), 'actions', { page: this.get('page') }).then(
+        this.get('currentUser.user').then(user => this.get('store').queryHasMany(user, 'actions', { page }).then(
             response => this._setPageProperties(response)
             , () => {
                 // Error
-                this.set('listLoading', false);
-                this.get('toast').error(this.get('errorMessage'));
+                this._handleLoadError();
             },
-        );
+        ));
+    },
+
+    _handleLoadError() {
+        this.set('listLoading', false);
+        this.set('loadingMore', false);
+        this.get('toast').error(this.get('errorMessage'));
     },
 
     _setPageProperties(response) {
-        if (this.get('loadingPage') === this.get('page')) {
-            this.setProperties({
-                actionsList: this.get('actionsList').concat(response.toArray()),
-                totalPages: Math.ceil(response.get('links.meta.total') / response.get('links.meta.per_page')),
-                listLoading: false,
-            });
-        }
+        this.setProperties({
+            actionsList: this.get('actionsList').concat(response.toArray()),
+            totalPages: Math.ceil(response.get('links.meta.total') / response.get('links.meta.per_page')),
+            listLoading: false,
+            loadingMore: false,
+        });
     },
 
     nextPage() {
         this.incrementProperty('page');
+        this.set('loadingMore', true);
         this.loadPage();
     },
 });
--- a/app/components/action-feed/component.js
+++ b/app/components/action-feed/component.js
@@ -13,10 +13,42 @@ import { translationMacro as t } from 'ember-i18n';
 export default Ember.Component.extend({
     store: Ember.inject.service(),
     toast: Ember.inject.service(),
+    currentUser: Ember.inject.service(),
 
     classNames: ['action-feed'],
 
     errorMessage: t('components.action-feed.error_loading'),
+
+    page: 1,
+
+    init() {
+        this._super(...arguments);
+        this.actionsList = [];
+        this.loadPage();
+    },
+
+    loadPage() {
+        const page = this.get('page');
+        this.set('loadingPage', page);
+        this.get('user').query('actions', {page: this.get('page')}).then((response) => {
+            if (this.get('loadingPage') === page) {
+                this.setProperties({
+                    actionsList: this.get('actionsList').concat(response.toArray()),
+                    totalPages: response.get('meta.total'),
+                    loadingPage: null,
+                });
+            }
+        }, () => {
+            // Error
+            this.set('loadingPage', null);
+            this.get('toast').error(this.get('errorMessage'));
+        });
+    },
+
+    nextPage() {
+        this.incrementProperty('page');
+        this.loadPage();
+    },
 
     moreActions: Ember.computed('totalPages', 'page', function() {
         return this.get('page') < this.get('totalPages');
--- a/app/components/action-feed/component.js
+++ b/app/components/action-feed/component.js
@@ -48,8 +48,8 @@ export default Component.extend({
     },
 
     _setPageProperties(response) {
+        this.get('actionsList').pushObjects(response.toArray());
         this.setProperties({
-            actionsList: this.get('actionsList').pushObjects(response.toArray()),
             totalPages: Math.ceil(response.get('links.meta.total') / response.get('links.meta.per_page')),
             listLoading: false,
             loadingMore: false,
--- a/app/components/action-feed/component.js
+++ b/app/components/action-feed/component.js
@@ -31,18 +31,15 @@ export default Component.extend({
         this._super(...arguments);
         this.loadPage();
         this.set('listLoading', true);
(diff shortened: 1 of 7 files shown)


app/routes/index.js
```diff
@@ -11,15 +11,15 @@ export default Route.extend({
         page: { refreshModel: true },
```
aaxelb: Could you remove this, too?

app/routes/index.js
```diff
@@ -11,15 +11,15 @@ export default Route.extend({
...
+    model() {
         if (!this.get('session.isAuthenticated')) {
             return this._emptyModels();
         }
         return hash({
             providers: this.get('store').query('preprint-provider', {
                 'filter[permissions]': 'view_actions,set_up_moderation',
             }),
-            actionsList: this.get('currentUser.user').then(this._getActions.bind(this, params.page)),
+            user: this.get('currentUser.user'),
```
aaxelb: Instead of getting the user here, why not get `currentUser.user` in the component? Then the model could be just a list of providers instead of a hash, which would be nice.

app/components/action-feed/component.js
```diff
@@ -12,15 +12,54 @@ import { translationMacro as t } from 'ember-i18n';
...
+        this.actionsList = [];
+        this.loadPage();
+        this.set('listLoading', true);
+        this.actionsList = [];
+    },
+
+    loadPage() {
+        const page = this.get('page');
+        this.set('loadingPage', page);
+        this.get('store').queryHasMany(this.get('user'), 'actions', { page: this.get('page') }).then(
```
aaxelb: You already got `page`, so the query params could be simply `{ page }`.

app/components/action-feed/component.js
```diff
@@ -12,15 +12,54 @@ import { translationMacro as t } from 'ember-i18n';
...
+            response => this._setPageProperties(response)
+            , () => {
+                // Error
```
aaxelb: Could you put this in a named method, like you did with `_setPageProperties`?

app/components/action-feed/component.js
```diff
@@ -12,15 +12,54 @@ import { translationMacro as t } from 'ember-i18n';
...
     moreActions: computed('totalPages', 'page', function() {
         return this.get('page') < this.get('totalPages');
     }),
+
+    init() {
+        this._super(...arguments);
+        this.actionsList = [];
+        this.loadPage();
+        this.set('listLoading', true);
+        this.actionsList = [];
```
aaxelb: `actionsList` is initialized twice

app/components/action-feed/template.hbs
```diff
@@ -1,4 +1,4 @@
-{{#if actionsList.isPending}}
+{{#if listLoading}}
```
aaxelb: Instead of hiding the list while it's loading the next page, I think it'd be nicer to simply hide the "See more" link and show the loading indicator below the list of actions.

app/components/action-feed/component.js
```diff
@@ -12,15 +12,54 @@ import { translationMacro as t } from 'ember-i18n';
...
+        if (this.get('loadingPage') === this.get('page')) {
+            this.setProperties({
+                actionsList: this.get('actionsList').concat(response.toArray()),
+                totalPages: Math.ceil(response.get('links.meta.total') / response.get('links.meta.per_page')),
+                listLoading: false,
+            });
+        }
+    },
+
+    nextPage() {
```
aaxelb: I think this should be in an `actions` hash?

app/components/action-feed/component.js
```diff
@@ -12,15 +12,54 @@ import { translationMacro as t } from 'ember-i18n';
...
 
     moreActions: computed('totalPages', 'page', function() {
         return this.get('page') < this.get('totalPages');
     }),
+
+    init() {
+        this._super(...arguments);
+        this.actionsList = [];
+        this.loadPage();
+        this.set('listLoading', true);
```
aaxelb: Seems like `listLoading` and `loadingPage` could be combined -- check whether `loadingPage` is truthy in the template and reset it to `null` when it's done loading.

app/components/action-feed/component.js
```diff
@@ -12,15 +12,54 @@ import { translationMacro as t } from 'ember-i18n';
...
+            , () => {
+                // Error
+                this.set('listLoading', false);
+                this.get('toast').error(this.get('errorMessage'));
+            },
+        );
+    },
+
+    _setPageProperties(response) {
+        if (this.get('loadingPage') === this.get('page')) {
```
aaxelb: Will this ever be false?

aaxelb (changes requested):

app/components/action-feed/template.hbs
sheriefvt: this is handled now..thanks 👍

app/components/action-feed/component.js
```diff
@@ -12,15 +12,57 @@ import { translationMacro as t } from 'ember-i18n';
...
+        this.loadPage();
+        this.set('listLoading', true);
+        this.actionsList = [];
+    },
+
+    loadPage() {
+        const page = this.get('page');
+        this.get('currentUser.user').then(user => this.get('store').queryHasMany(user, 'actions', { page }).then(
+            response => this._setPageProperties(response)
+            , () => {
```
aaxelb: Could you move this comma to the line above, and get rid of the brackets? `() => this._handleLoadError()` would be more consistent with the line above.

app/routes/index.js
```diff
@@ -7,19 +6,13 @@ export default Route.extend({
...
             return this._emptyModels();
         }
-        return hash({
-            providers: this.get('store').query('preprint-provider', {
-                'filter[permissions]': 'view_actions,set_up_moderation',
-            }),
-            actionsList: this.get('currentUser.user').then(this._getActions.bind(this, params.page)),
+        return this.get('store').query('preprint-provider', {
+            'filter[permissions]': 'view_actions,set_up_moderation',
         }).catch(this._emptyModels.bind(this)); // On any error, assume no permissions to anything.
```
aaxelb, line 16: `_emptyModels` still returns `{ providers: [], actionsList: [] }`. I think you could just remove it and use `[]` in its place.

app/components/action-feed/template.hbs
```diff
@@ -8,11 +8,19 @@
...
-    {{#if moreActions}}
+    {{#if (and moreActions (not loadingMore))}}
         <div class='see-more'>
             <a role='button' {{action nextPage}}>{{t 'components.action-feed.see_more'}}</a>
         </div>
     {{else if (eq actionsList.length 0)}}
         <div class="text-center">{{t 'components.action-feed.no_actions'}}</div>
     {{/if}}
+    {{#if loadingMore}}
+        <div class='loading'>
```
aaxelb: I think it'd be better to use `{{loading-indicator}}`, for consistency with the rest of the app.

app/components/action-feed/template.hbs
```diff
@@ -8,11 +8,19 @@
     {{#each actionsList key='id' as |action|}}
         {{action-feed-entry action=action toDetail=toDetail}}
     {{/each}}
-    {{#if moreActions}}
+    {{#if (and moreActions (not loadingMore))}}
         <div class='see-more'>
             <a role='button' {{action nextPage}}>{{t 'components.action-feed.see_more'}}</a>
         </div>
     {{else if (eq actionsList.length 0)}}
```
aaxelb: This seems like a perfect opportunity for an each/else block, like
```
{{#each actionsList ...}}
  {{action-feed-entry ...}}
{{else}}
  No actions...
{{/each}}
```

app/components/action-feed/template.hbs
```diff
@@ -8,11 +8,19 @@
     {{#each actionsList key='id' as |action|}}
         {{action-feed-entry action=action toDetail=toDetail}}
     {{/each}}
-    {{#if moreActions}}
+    {{#if (and moreActions (not loadingMore))}}
         <div class='see-more'>
```
aaxelb: When I click on "See more", the "Send us your thoughts" banner jumps down while the loading indicator is displayed. Make sure the "See more" and the loading indicator div use the same amount of vertical space.

app/components/action-feed/component.js
```diff
@@ -12,15 +12,57 @@ import { translationMacro as t } from 'ember-i18n';
...
+
+    _handleLoadError() {
+        this.set('listLoading', false);
+        this.set('loadingMore', false);
+        this.get('toast').error(this.get('errorMessage'));
+    },
+
+    _setPageProperties(response) {
+        this.setProperties({
+            actionsList: this.get('actionsList').concat(response.toArray()),
```
aaxelb: I think it would make more sense to append to `actionsList` using `pushObjects`, rather than creating and assigning a new array.

app/components/action-feed/component.js
```diff
@@ -12,15 +12,57 @@ import { translationMacro as t } from 'ember-i18n';
...
 
     errorMessage: t('components.action-feed.error_loading'),
 
     moreActions: computed('totalPages', 'page', function() {
         return this.get('page') < this.get('totalPages');
     }),
+
+    init() {
+        this._super(...arguments);
+        this.loadPage();
```
aaxelb: Minor: It probably wouldn't ever cause a problem, since the promises will take time to resolve, but it seems wrong to call `loadPage` before initializing the other properties.

app/components/action-feed/component.js
```diff
@@ -12,15 +12,57 @@ import { translationMacro as t } from 'ember-i18n';
...
+        this.set('listLoading', true);
+        this.actionsList = [];
```
aaxelb: use `this.set`

app/components/action-feed/component.js
```diff
@@ -12,15 +12,57 @@ import { translationMacro as t } from 'ember-i18n';
...
  */
+
 export default Component.extend({
     store: service(),
     toast: service(),
-
+    currentUser: service(),
     classNames: ['action-feed'],
+    page: 1,
+    listLoading: false,
```
aaxelb: Minor: Why is this `false` if it's always re-initialized in `init`?

app/components/action-feed/component.js
```diff
@@ -12,15 +12,57 @@ import { translationMacro as t } from 'ember-i18n';
...
+    init() {
+        this._super(...arguments);
+        this.loadPage();
+        this.set('listLoading', true);
+        this.actionsList = [];
+    },
+
+    loadPage() {
+        const page = this.get('page');
+        this.get('currentUser.user').then(user => this.get('store').queryHasMany(user, 'actions', { page }).then(
```
aaxelb: I think it'd be cleaner to chain these promises at the same level, like
```
this.get('currentUser.user')
    .then(user => store.queryHasMany(...))
    .then(this._setPageProperties.bind(this))
    .catch(this._handleLoadError.bind(this));
```
(using `.bind` vs `() =>` doesn't matter, but I like binding because you don't need to repeat the arguments that are being passed)

aaxelb (changes requested):

app/components/action-feed/template.hbs
```diff
@@ -1,18 +1,16 @@
...
+{{#if listLoading}}
+    {{loading-indicator classes='ball-dark text-center loading'}}
 {{else}}
     {{#each actionsList key='id' as |action|}}
         {{action-feed-entry action=action toDetail=toDetail}}
+    {{else}}
+        <div class="text-center">{{t 'components.action-feed.no_actions'}}</div>
     {{/each}}
-    {{#if moreActions}}
+    {{#if (and moreActions (not loadingMore))}}
```
aaxelb: This logic will display the loading indicator if `!moreActions`, which is wrong (go to the last page of actions, the loading indicator stays there). Seems like it should be something like:
```hbs
{{#if loadingMore}}
  {{loading-indicator}}
{{else if moreActions}}
  See more
{{/if}}
```

app/routes/index.js
```diff
@@ -28,9 +21,6 @@ export default Route.extend({
     },
 
     _emptyModels() {
-        return {
-            providers: [],
-            actionsList: [],
-        };
+        return [];
```
aaxelb, line 24: There's no point to a method that just returns `[]`, please remove `_emptyModels` and use a literal `[]` instead.

app/components/action-feed/component.js
```diff
@@ -12,15 +12,53 @@ import { translationMacro as t } from 'ember-i18n';
...
+
+    _handleLoadError() {
+        this.set('listLoading', false);
+        this.set('loadingMore', false);
+        this.get('toast').error(this.get('errorMessage'));
+    },
+
+    _setPageProperties(response) {
+        this.setProperties({
+            actionsList: this.get('actionsList').pushObjects(response.toArray()),
```
aaxelb: `pushObjects` modifies the array in place, this is just setting `actionsList = this.get('actionsList')`. It'd make more sense to do that one line up:
```js
this.get('actionsList').pushObjects(...);
this.setProperties({
    totalPages: ...
});

aaxelb (changes requested):
Just a few small things

aaxelb (approved):

PR #73, opened 2017-10-23
[MOD-136][Improvement] Loading for the dashboard

laurenbarker:
## TODO
- [x] Use the cos-fork of ember-content-placeholders

## Changes
* Install ember-concurrency
* Move loading data out of routes
* Add skeleton screens for dashboard

dashboard-skeleton


app/components/action-feed/template.hbs
```diff
@@ -1,14 +1,18 @@
-{{#if listLoading}}
-    {{loading-indicator classes='ball-dark text-center loading'}}
+{{#if (and loadActions.isRunning (not loadingMore))}}
```
aaxelb: I think we could get rid of `loadingMore` and DRY this up a little...
```hbs
{{#each actionsList ...}}
  entry
{{else if (not loadActions.isRunning)}}
  No actions
{{/if}}
{{#if loadActions.isRunning}}
  dummy list
{{else if moreActions}}
  see more
{{/if}}
```

app/controllers/index.js
```diff
@@ -35,4 +39,16 @@ export default Controller.extend({
...
         },
     },
+
+    init() {
+        this._super(...arguments);
+        this.get('fetchData').perform();
+    },
+
+    fetchData: task(function* () {
+        yield timeout(1000);
```
aaxelb: Is this left over from debugging?

app/routes/index.js
```diff
@@ -1,26 +1,4 @@
-import { inject as service } from '@ember/service';
 import Route from '@ember/routing/route';
 
 export default Route.extend({
```
aaxelb: Can we delete this file?

app/components/action-feed/component.js
```diff
@@ -22,43 +23,41 @@ export default Component.extend({
...
             loadingMore: false,
         });
+        this.get('loadActions').perform();
     },
 
+    loadActions: task(function* () {
+        const page = this.get('page');
+        try {
+            const results = yield this.get('currentUser.user')
+                .then(user => this.get('store').queryHasMany(user, 'actions', { page }));
```
aaxelb: This isn't directly related to your changes, but would it load faster if we added `embed: 'target'` here?

app/components/action-feed/component.js
```diff
@@ -22,43 +23,41 @@ export default Component.extend({
...
+            actionsList: [],
             loadingMore: false,
         });
+        this.get('loadActions').perform();
     },
 
+    loadActions: task(function* () {
+        const page = this.get('page');
+        try {
+            const results = yield this.get('currentUser.user')
```
aaxelb: Thoughts on the ember-concurrency idiom instead of chaining promises?
```js
const user = yield this.get('currentUser.user');
const actions = yield this.get('store').queryHasMany(user, ...);
this.get('actionsList').pushObjects(actions.toArray());
...
```

app/components/action-feed/component.js
```diff
@@ -22,43 +23,41 @@ export default Component.extend({
...
+        } catch (e) {
+            this.get('toast').error(this.get('errorMessage'));
+        }
+    }),
+
     nextPage() {
         this.incrementProperty('page');
         this.set('loadingMore', true);
-        this.loadPage();
+        this.get('loadActions').perform();
```
aaxelb: If you add `.drop()` to `loadActions`, seems like we could safely increment `page` inside it and get rid of `nextPage`

app/components/action-feed-entry/style.scss
```diff
@@ -1,10 +1,6 @@
 & {
   clear: both;
 
-  .entry-body {
-    padding: 20px 15px;
```
aaxelb, line 5: Why no padding?
screen shot 2017-11-06 at 4 46 51 pm

aaxelb (changes requested):
Overall, looks awesome, just a smattering of small things.

Tiny bug: I think something happened moving data out of the route -- when I refresh the settings page, the count on the moderation tab is initially correct, then changes to `0` after a second.

...and there were some tests merged earlier today that this breaks, sorry about that.

app/components/action-feed-entry/style.scss
laurenbarker: Aw...I forgot about the hover box. The top of the list looks out of line with the provider box with top padding.

app/components/action-feed/component.js
laurenbarker: Embedding isn't faster but it prevents the piecemeal loading.

laurenbarker:
I don't know why the build isn't running. It passed on my branch...

app/components/action-feed-entry-skeleton/template.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="p-b-lg">
+    {{#content-placeholders as |placeholder|}}
+        <div class="icon--skeleton">{{placeholder.icon}}</div>
+    {{/content-placeholders}}
+    <div class="action-body--skeleton">
+        {{#content-placeholders as |placeholder|}}
```
aaxelb: Could these two `{{#content-placeholders as |placeholder|}}` blocks be combined in one?

app/components/landing-page/component.js
```diff
@@ -0,0 +1,55 @@
...
+        });
+        this.get('fetchData').perform();
+    },
+
+    actions: {
+        transitionToDetail(provider, reviewable) {
+            this.transitionToRoute('preprints.provider.preprint-detail', provider.get('id'), reviewable.get('id'));
+        },
+        setupProvider(provider) {
+            this.transitionToRoute('preprints.provider.setup', provider.id);
```
aaxelb: `Uncaught TypeError: this.transitionToRoute is not a function`

The `router` service would help.

app/components/landing-page/template.hbs
```diff
@@ -0,0 +1,47 @@
+{{#if fetchData.isRunning}}
+    {{loading-indicator-page}}
+{{else}}
+    {{#if showDashboard}}
+        <div class='reviews-dashboard content'>
```
aaxelb: If we're doing all components, I think it'd be nice to make the landing page and dashboard separate components. Maybe the logic to switch between them would make sense on the controller? Not super important, but now seems like a convenient time to clean this up, if you don't mind...

app/components/action-feed/component.js
```diff
@@ -18,47 +19,43 @@ export default Component.extend({
...
-        this.get('toast').error(this.get('errorMessage'));
-    },
-
-    _setPageProperties(response) {
-        this.get('actionsList').pushObjects(response.toArray());
         this.setProperties({
-            totalPages: Math.ceil(response.get('links.meta.total') / response.get('links.meta.per_page')),
-            listLoading: false,
+            actionsList: [],
             loadingMore: false,
```
aaxelb: `loadingMore` is gone

aaxelb (changes requested):
Found a little bug and had a few suggestions.

app/components/landing-page/template.hbs
laurenbarker: The request for providers is what messes up the other pages. So I can move some stuff around but I don't think the logic to switch between them can go in the controller since it's called when the app refreshes no matter what page the user is on (that was the bug on the settings page).

app/components/reviews-dashboard/template.hbs
```diff
@@ -0,0 +1,23 @@
...
+        </div>
+    </div>
+    <div class='container'>
+        <div class='row'>
+            <div class='col-md-4 col-md-push-8'>
+                {{dashboard-sidebar providers=providers}}
+            </div>
+            <div class='col-md-8 col-md-pull-4'>
+                {{action-feed
+                    toDetail=(action transitionToDetail)
```
aaxelb, line 18: Since `transitionToDetail` is already a closure action, I think you could just do `toDetail=transitionToDetail`. But more than that, it feels wrong to pass this action so far down (controller => reviews-dashboard => action-feed => action-feed-entry) when action-feed-entry could just use the router service and trigger a transition all by itself.

app/components/landing-page/template.hbs
```diff
@@ -0,0 +1,12 @@
+<div class="reviews-header">
+    <div class="container">
+        <div class="row">
+            <div class="text-center m-t-lg col-xs-12">
+                <div class="m-b-lg reviews-brand"></div>
+                {{#if showSetup}}
+                    {{start-moderating action=(action setupProvider) providers=providersToSetup}}
```
aaxelb, line 7: Same thing here: It seems cleaner for start-moderating to use the router service and trigger the transition itself, rather than passing `setupProvider` all the way down from the controller.

aaxelb (changes requested):
Just one last thing, if it makes sense to you.

app/components/reviews-dashboard/template.hbs
laurenbarker: I was trying to avoid that since it seems very anti-Ember but we are ignoring pretty much everything they recommend so I guess it doesn't really matter. I do hate passing things down 10 levels.

aaxelb: I think it'd make more sense if these components were meant to be reusable outside this app, or the action varied depending on other state, or it did anything other than a simple transition.

They added `router`, though, so I figure we may as well use it.

laurenbarker: @aaxelb, bad news. We can't use the router service yet because their is a [bug](https://github.com/emberjs/ember.js/pull/15613) that breaks when all query params aren't present. So when we try and use `this.get('router').transitionTo(...)` we always get an error about the [ticket query parameter on the application route](https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/mixins/osf-cookie-login-controller.js#L20).
```
"Assertion Failed: You passed the `false` query parameter during a transition into application, please update to ticket"
``` 

Fix release:
> Apologies for not getting this in for 2.16, I'll get the changes into the next 2.17 beta release and once some folks can confirm all is working well for us I'll pull down into 2.16 for the LTS.

aaxelb: That fix is in [2.16.1](https://github.com/emberjs/ember.js/releases/tag/v2.16.1)... we should upgrade. In another ticket.

aaxelb (approved):

app/components/reviews-dashboard/template.hbs
laurenbarker: Oh nice, I missed that. I'll add a ticket 👍

PR #74, opened 2017-10-23
[Improvement][MOD-77] Add length validation to moderator comments

sheriefvt:
## Problem
Users can submit comments that can break backend if exceeds a certain limit.  

## Solution 

- Add limit on the ember front end with error message to make limit obvious when reached
- Set limit at 65535 characters

## Demo
Note: limit was set to 10 for demo only.

## Ticket
https://openscience.atlassian.net/browse/MOD-77


app/components/preprint-status-banner/component.js
```diff
@@ -161,6 +164,17 @@ export default Component.extend({
         return isBlank(this.get('reviewerComment'));
     }),
 
+    commentExceedsLimit: computed('reviewerComment', function() {
```
aaxelb: Simpler: `computed.gt('reviewerComment.length', COMMENT_LIMIT)`

app/components/preprint-status-banner/component.js
```diff
@@ -161,6 +164,17 @@ export default Component.extend({
...
+        return this.get('reviewerComment.length') > COMMENT_LIMIT;
+    }),
+
+    commentLengthErrorMessage: computed(function () {
```
aaxelb: This might be cleaner with the [`t` computed property macro](https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Translating-Text#translation-computed-property-macro)

app/components/preprint-status-banner/style.scss
```diff
@@ -127,6 +127,10 @@
             margin: 10px 10px 5px 10px;
           }
         }
+
+        .comment-length-error {
+         color: red;
```
aaxelb: indentation

aaxelb (changes requested):
Just a few small things.

Also, if we let them type more than the limit, what would you think about displaying the comment's current length so they know how much they need to delete? Or, what about setting `maxlength` on the textarea so they can't type over the limit (in most browsers...)?

app/components/preprint-status-banner/style.scss
```diff
@@ -127,6 +127,10 @@
             margin: 10px 10px 5px 10px;
           }
         }
+
+        .comment-length-error {
+          color: red;
```
aaxelb: Prefer using a variable in `app/styles/_variables.scss` for colors.

app/locales/en/translations.js
```diff
@@ -236,6 +236,7 @@ export default {
                     update_comment: 'Update comment',
                 },
                 comment_placeholder: 'Explain the reasoning behind your decision (optional)',
+                comment_length_error: 'Comment exceeded maximum limit of {{limit}} characters by {{difference}}.',
```
aaxelb: This seems okay to me, but @laurenbarker, thoughts on this language/approach?

aaxelb (changes requested):
Use variables for colors, otherwise looks pretty good.

app/components/preprint-status-banner/style.scss
laurenbarker: `$state-danger-text` is the color that the preprint form uses for validation errors.

app/locales/en/translations.js
laurenbarker: The language used for the preprint form is "Abstract is too short (minimum is 20 characters)" so we could do "Comment is too long (maximum is {{limit}})" for consistency.

aaxelb: Should it also say how far over it is? If we're letting users type beyond the limit, seems like it'd be good to let them know how much they need to delete.

laurenbarker: How about "Comment is {{difference}} characters too long (maximum is {{limit}})"?

aaxelb (approved):

PR #76, opened 2017-10-25
[Improvement][MOD-159] Add DOI message - Reviews

sheriefvt:
## Problem
Users don't know when they will have a DOI for their preprint.

## Solution
Add "DOI created after moderator approval." on pending pre-moderation preprints

image
image

## Ticket
https://openscience.atlassian.net/browse/MOD-159

### DIFF
--- a/app/templates/preprints/provider/preprint-detail.hbs
+++ b/app/templates/preprints/provider/preprint-detail.hbs
@@ -68,11 +68,10 @@
                 </div>
 
                 <section class="p-t-xs">
-                    {{theme.provider.reviewsWorkflow}}
                     {{#if model.preprintDoiUrl}}
                         <h4 class="p-v-md f-w-md"><strong>{{t "content.preprint_doi"}}</strong></h4>
                         <a href={{model.preprintDoiUrl}}>{{extract-doi model.preprintDoiUrl}}</a>
-                    {{else if (eq theme.provider.reviewsWorkflow 'pre-moderation')}}
+                    {{else if (not model.isPublished)}}
                         <h4 class="p-v-md f-w-md"><strong>{{t "content.preprint_doi"}}</strong></h4>
                         {{t 'content.preprint_pending_doi'}}
                     {{/if}}
--- a/app/templates/preprints/provider/preprint-detail.hbs
+++ b/app/templates/preprints/provider/preprint-detail.hbs
@@ -67,12 +67,16 @@
                     </button>
                 </div>
 
-                {{#if model.preprintDoiUrl}}
-                    <section class="p-t-xs">
+                <section class="p-t-xs">
+                    {{theme.provider.reviewsWorkflow}}
+                    {{#if model.preprintDoiUrl}}
                         <h4 class="p-v-md f-w-md"><strong>{{t "content.preprint_doi"}}</strong></h4>
                         <a href={{model.preprintDoiUrl}}>{{extract-doi model.preprintDoiUrl}}</a>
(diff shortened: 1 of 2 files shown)


app/templates/preprints/provider/preprint-detail.hbs
```diff
@@ -67,12 +67,16 @@
                     </button>
                 </div>
 
-                {{#if model.preprintDoiUrl}}
-                    <section class="p-t-xs">
+                <section class="p-t-xs">
+                    {{theme.provider.reviewsWorkflow}}
```
aaxelb: Delete this

app/templates/preprints/provider/preprint-detail.hbs
```diff
@@ -67,12 +67,16 @@
...
+                    {{#if model.preprintDoiUrl}}
                         <h4 class="p-v-md f-w-md"><strong>{{t "content.preprint_doi"}}</strong></h4>
                         <a href={{model.preprintDoiUrl}}>{{extract-doi model.preprintDoiUrl}}</a>
-                    </section>
-                {{/if}}
+                    {{else if (eq theme.provider.reviewsWorkflow 'pre-moderation')}}
```
aaxelb: I think it'd be better to be more generic: display the message if `model.isPublished` is false. No matter the workflow, it will get a DOI when it transitions into a public state.

aaxelb (changes requested):

app/templates/preprints/provider/preprint-detail.hbs
laurenbarker: We would need a different message for that because "after moderator approval" would be inaccurate for post-moderation or no moderation.

Could be "DOI created after preprint is made public".

aaxelb: I'd be good with that.

Though for post-moderation, the only way `isPublished` would be false is if it's rejected, so it would require explicit approval. And with no moderation, this screen isn't even accessible.

laurenbarker: Rejected, post-moderation preprints would have a DOI. You can still see the screen if you make your preprint if you make your node private though.

aaxelb: Yeah, I assumed this would still be an `else if`, so just for preprints with no DOI. Once https://github.com/CenterForOpenScience/osf.io/pull/7837 is merged, we might be able to simplify and display this message whenever there's no DOI, which would be nice.

(If you make your node private, this page redirects to `page-not-found`. Or at least it should...)

laurenbarker: It redirects to `page-not-found` for non-contributors.
screen shot 2017-10-25 at 3 21 03 pm

laurenbarker: Whoops, this is my bad. I thought this change was on Preprints not Reviews. Disregard pretty much all of my comments.

sheriefvt: @laurenbarker @aaxelb what changes do u suggest, and do you suggest change the wording

aaxelb: ~~Change this line to `else if !model.isPublished` (or equivalent), and change wording to "DOI created after preprint is made public". Does that sound right @laurenbarker?~~
@sheriefvt Updated notes added to the ticket

aaxelb (approved):

PR #77, opened 2017-10-25
[MOD-163][Fix] Don't disable the default for radio buttons

laurenbarker:
## Purpose
The setup form would not work if you click on the actual circle for the radio buttons instead of the whole block.

## Changes
Don't prevent the default for radio buttons.


aaxelb (approved):
Who actually clicks on the radio button, though?

laurenbarker:
Our QA team 🥇

PR #79, opened 2017-10-27
[Fix] Fix see more and loading-indicator

sheriefvt:
Clicking on "see more" link button on the dashboard is not showing the loading-indicator

### DIFF
--- a/app/components/action-feed/component.js
+++ b/app/components/action-feed/component.js
@@ -23,7 +23,7 @@ export default Component.extend({
     errorMessage: t('components.action-feed.error_loading'),
 
     moreActions: computed('totalPages', 'page', 'loadingMore', function() {
-        return this.get('loadingMore') ?  false : this.get('page') < this.get('totalPages');
+        return this.get('loadingMore') ? false : this.get('page') < this.get('totalPages');
     }),
 
     init() {
--- a/app/components/action-feed/component.js
+++ b/app/components/action-feed/component.js
@@ -22,8 +22,8 @@ export default Component.extend({
 
     errorMessage: t('components.action-feed.error_loading'),
 
-    moreActions: computed('totalPages', 'page', function() {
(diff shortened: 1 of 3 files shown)


app/components/action-feed/component.js
```diff
@@ -22,8 +22,8 @@ export default Component.extend({
 
     errorMessage: t('components.action-feed.error_loading'),
 
-    moreActions: computed('totalPages', 'page', function() {
-        return this.get('page') < this.get('totalPages');
+    moreActions: computed('totalPages', 'page', 'loadingMore', function() {
+        return this.get('loadingMore') ? false : this.get('page') < this.get('totalPages');
```
laurenbarker: When is `this.get('page') < this.get('totalPages')`? Would it be covered by `loadingMore`?

laurenbarker: Oh wait this makes sense 👍

laurenbarker (commented):
What if instead of `{{#if moreActions}}` and adding `loadingMore` to `moreActions` you did `{{#if (and moreActions (not loadingMore))}}`? I'm not totally sure that works but it seems weird that `moreActions` would be false when it's loading.

sheriefvt:
@laurenbarker tested both works, shall i do this change or keep as it is.

laurenbarker:
Thanks for checking! @aaxelb, your call.

aaxelb:
I think the current (updated) logic makes sense, but could be simplified to the equivalent:
```hbs
{{#if loadingMore}}
    {{loading-indicator ...}}
{{else if moreActions}}
    See more
{{/if}}
```

aaxelb:
Also, please add a line to the changelog, under `Fixed`

sheriefvt:
@laurenbarker @aaxelb done 🐼

aaxelb (approved):

PR #57, opened 2017-10-06
[Feature][MOD-115] Show moderator name on accepted/rejected records

sheriefvt:
## Problem
For MVP we weren't able to access the moderator's name who made the most recent decision on the moderation list page. It's better to display the relevant information.

## Changes 
Update the text on the rejected/accepted list view to: `DECISION on DATE by MODERATOR` 

image
image

## Ticket
https://openscience.atlassian.net/browse/MOD-115

## Note
HOLD-landing-page template removed on this [commit](https://github.com/CenterForOpenScience/ember-osf-reviews/pull/57/commits/726b61d794ec8ffc1640d283d5557c41cb17a2d7)

PR #84, opened 2017-11-01
[MOD-174][Hotfix] Fix moderator seeing old comment bug

laurenbarker:
## Purpose
Fix the bug where moderators wouldn't see the most recent comment until they refreshed the page.

## Changes
Update `init` on preprint detail page to get the most recent action.

## Side Effects
None expected

### DIFF
--- a/app/components/preprint-status-banner/component.js
+++ b/app/components/preprint-status-banner/component.js
@@ -147,14 +147,7 @@ export default Component.extend({
     }),
 
     latestAction: computed('submission.actions.[]', function() {
-        if (!this.get('submission.actions.length')) {
-            return null;
-        }
-        // on create, Ember puts the new object at the end of the array
-        // https://stackoverflow.com/questions/15210249/ember-data-insert-new-item-at-beginning-of-array-instead-of-at-the-end
-        const first = this.get('submission.actions.firstObject');
-        const last = this.get('submission.actions.lastObject');
-        return moment(first.get('dateModified')) > moment(last.get('dateModified')) ? first : last;
+        return this._getRecentAction.bind(this.get('submission.actions'));
     }),
 
     noComment: computed('reviewerComment', function() {
@@ -229,7 +222,9 @@ export default Component.extend({
     }),
 
     init() {
-        this.get('submission.actions').then(this._handleActions.bind(this));
+        this.get('submission.actions')
+            .then(this._getRecentAction.bind(this))
+            .then(this._handleActions.bind(this));
         return this._super(...arguments);
     },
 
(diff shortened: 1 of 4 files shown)


aaxelb:
If the action was just created, it should be available somewhere in `model.actions`, and reloading should be unnecessary. Maybe there's a bug in the `latestAction` logic, or it's not recomputed for some reason?

aaxelb:
...or the [method that sets the comment value](https://github.com/CenterForOpenScience/ember-osf-reviews/blob/develop/app/components/preprint-status-banner/component.js#L271) always grabs `firstObject` instead of the latest.

laurenbarker:
Yeah, it's the init method. I'm just trying to figure out how to fix it now.

app/components/preprint-status-banner/component.js
```diff
@@ -251,10 +246,23 @@ export default Component.extend({
         },
     },
 
-    _handleActions(actions) {
-        if (actions.length) {
+    _getRecentAction(actions) {
```
aaxelb: To minimize conflicts with [Sherif's PR](https://github.com/CenterForOpenScience/ember-osf-reviews/pull/57/files#diff-210068c358e26db48f45c5f18ae2d194), could you move this to a util?

app/components/preprint-status-banner/component.js
```diff
@@ -147,14 +147,7 @@ export default Component.extend({
...
     latestAction: computed('submission.actions.[]', function() {
-        if (!this.get('submission.actions.length')) {
-            return null;
-        }
-        // on create, Ember puts the new object at the end of the array
-        // https://stackoverflow.com/questions/15210249/ember-data-insert-new-item-at-beginning-of-array-instead-of-at-the-end
-        const first = this.get('submission.actions.firstObject');
-        const last = this.get('submission.actions.lastObject');
-        return moment(first.get('dateModified')) > moment(last.get('dateModified')) ? first : last;
+        return this._getRecentAction.bind(this.get('submission.actions'));
```
aaxelb: This seems wrong, I don't think you meant to include `.bind`?

aaxelb (changes requested):
Just a couple small things

app/components/preprint-status-banner/component.js
laurenbarker: Should I just copy his code?

aaxelb: If you want to. Or write it however you like -- since this is a hotfix, it'll get in first and @sheriefvt can use or modify it however it makes sense.

laurenbarker: Ah, I see what you mean.

app/components/preprint-status-banner/component.js
```diff
@@ -229,7 +222,9 @@ export default Component.extend({
     }),
 
     init() {
-        this.get('submission.actions').then(this._handleActions.bind(this));
+        this.get('submission.actions')
+            .then(this._getRecentAction.bind(this))
```
aaxelb: `_getRecentAction` seems unnecessary, could just be `.then(latestAction)`, I think.

aaxelb (approved):

PR #85, opened 2017-11-01
Mutli-stage docker build

binoculars:
## Purpose

Streamlines builds by using a common base image

## Summary of Changes

* Add yarnrc
* Updates Dockerfile
* Removes Dockerfile-alpine (no longer needed)
* Change default port back to 4200, can set from CLI and use docker port mapping

* Must update Jenkins build config immediately before merging (must also do this again prior to merging to master)

## Ticket

N/A


.ember-cli
```diff
@@ -1,4 +1,4 @@
 {
-  "port": 4400,
+  "port": 4200,
```
aaxelb, line 2: This was 4400 so it could peacefully coexist locally with preprints at 4200. Is that not a problem with this update?

binoculars: No, we have to override it in the OSF anyway. So for simplicity, leaving it as 4200 is better. I'll  have another PR that updates the osf.io docker-compose.yml file with the correct port numbers for each ember app. You can also use `yarn start --port 4400` if you are running it locally outside docker, or `docker run -it <image> yarn start --port 4400` inside docker.

aaxelb: Makes sense 👍

aaxelb (approved):

PR #86, opened 2017-11-01
[MOD-175][Hotfix] Include api version in queryHasMany request

laurenbarker:
## Purpose
Prevent timezone issues.

## Changes
* Included header with api version in queryHasMany request.
    * Used version 2.5 because I get errors with 2.6
* Updated accessing the metadata since it changed between API version.

## Side effects
None expected.

### DIFF
--- a/app/services/store.js
+++ b/app/services/store.js
@@ -27,6 +27,9 @@ export default Store.extend({
             if (url) {
                 $.ajax(url, {
                     data: queryParams,
+                    headers: {
+                        ACCEPT: 'application/vnd.api+json; version=2.5',
+                    },
                     xhrFields: {
                         withCredentials: true,
                     },
--- a/app/services/store.js
+++ b/app/services/store.js
@@ -19,6 +19,7 @@ export default Store.extend({
      * @returns {ArrayPromiseProxy} Promise-like array proxy, resolves to the records fetched
      */
     queryHasMany(model, propertyName, queryParams) {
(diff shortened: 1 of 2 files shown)


app/services/store.js
```diff
@@ -27,6 +27,9 @@ export default Store.extend({
             if (url) {
                 $.ajax(url, {
                     data: queryParams,
+                    headers: {
+                        ACCEPT: 'application/vnd.api+json; version=2.5',
```
aaxelb, line 32: Could you import `ember-osf/adapters/osf-adapter` and use its `headers` instead?

aaxelb (changes requested):
Out of curiosity, what errors did you get with 2.6?

laurenbarker:
500 from the API:
```
TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.
[01/Nov/2017 15:23:55] "GET /v2/preprint_providers/osf/preprints/?filter%5Breviews_state%5D=accepted&filter%5Bnode_is_public%5D=true&meta%5Breviews_state_counts%5D=true&sort=-date_last_transitioned&page=1 HTTP/1.1" 500 59
```

app/services/store.js
laurenbarker: Tried and failed to import the headers as discussed on flowdock 😞 . Added ticket and comment.

aaxelb (approved):

PR #88, opened 2017-11-02
[Improvement][MOD-76] Make the reviews app mobile friendly

sheriefvt:
## Problem
Adjust the Reviews styling/layout on small screens so that the reviews app is mobile friendly.

## Changes

- **Moderation List**

- **Preprint detail page**

- **Reviewer feedback dropdown**

- **Settings page**

## Ticket
https://openscience.atlassian.net/browse/MOD-76


app/components/moderation-list/style.scss
```diff
@@ -71,6 +72,29 @@
   }
 }
 
+@media screen and (max-width: $screen-medium-small) {
+  .moderation-list-heading {
+    .btn {
+      font-size: 14px;
```
aaxelb, line 78: It would be awesome to have this sort of responsive `font-size` for the entire app -- it looks weird when just the list heading shrinks and everything above and below it looks huge in comparison.

Would it be feasible to do something like that everywhere without repetitive code? Maybe consistently use `em`/`rem` for `font-size`, and a single media query on `html` to set `font-size` based on the viewport width? Or does that seem like a bad idea (or too much to fit in this ticket)?

app/components/preprint-status-banner/style.scss
```diff
@@ -135,3 +135,16 @@
     }
   }
 }
+@media (max-width: $screen-small-tablet) {
+  .recent-activity {
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
```
aaxelb: I'd rather not hide information on small screens. Could it be displayed on the next line instead?
First line: "Accepted", "Modify Decision" button
Second line: [Person] accepted this preprint on [Date]

app/styles/_variables.scss
```diff
@@ -45,7 +45,9 @@ $fa-font-path: 'fonts';
 $gradient-percentage: 15%;
 $nav-bar-height: 25px;
 $carousel-control-width: 10%;
-$screen-xs: 400px;
+$screen-xs: 420px;
+$screen-small-tablet: 700px;
+$screen-medium-small: 500px;
```
aaxelb, line 50: Can we use the [bootstrap breakpoints](https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_variables.scss#L286) instead of defining our own?

aaxelb (changes requested):
Looks good, this is definitely more usable on small screens. I had a few questions how we might make it better, mostly wondering if we could have responsive font sizes for everything, not just the moderation list heading.

app/components/preprint-status-banner/style.scss
sheriefvt: it could, but not guaranteed to fit in. Also, some author names might be too long for small screens.

app/styles/_variables.scss
sheriefvt: i tried but seems to look better at our custom breakpoints.

app/components/preprint-status-banner/style.scss
aaxelb: I think "not guaranteed to fit" is better than "guaranteed not to fit"

aaxelb: And if it were on its own row, it'd be okay to wrap to two lines.

sheriefvt: it might exceed two lines, do you want in this case to use ellipsis. The space in smaller screens like iphone 5 is limited

aaxelb: I don't know, I guess it depends how bad it looks. Generally speaking, we shouldn't hide information if we can avoid it.

sheriefvt: i will give it try

app/components/moderation-list/style.scss
sheriefvt: @laurenbarker do you have thoughts on this?

laurenbarker: It should probably be another ticket. Definitely worth doing though so I'll go ahead and add one.

aaxelb:
Not trying to nitpick, this ticket is mainly to get things functional on all screens, and you've done that well, but a few layout things stuck out that I'd like to fix.

Generally, on small screens the right and left margins should be equal. It looks weird when the right margin shrinks to zero while the left margin stays what it was.

### Status bar on the detail page
Make sure there's space between "recent activity" and the decision button (taken around ~700px)
screen shot 2017-11-08 at 2 57 12 pm

On smaller screens, make sure the space on the left and right are equal. Also maybe add some space above the "recent activity". 
screen shot 2017-11-08 at 2 58 48 pm

### Moderation list

Don't let the right margin go to zero, right and left should match.
screen shot 2017-11-08 at 3 20 09 pm

On the smallest screens, could you make the pending/accepted/rejected buttons stack vertically when they run out of horizontal space? Then the counts could stay visible (I don't like hiding information), and there would be space for the "sort" to live comfortably.
screen shot 2017-11-08 at 3 00 04 pm

----

Overall, though, looks good! I'm looking forward to having responsive font size (and padding/margin?) everywhere, but that's another ticket.

aaxelb:
Just one more nitpick and it's good to go!

The "Sort" control shouldn't leap around. When the screen narrows from this:
screen shot 2017-11-14 at 3 04 51 pm
to this:
screen shot 2017-11-14 at 3 04 05 pm
it should stay up at the top, with the same padding/margin from the top and right borders.

sheriefvt:
@aaxelb I feel it is better to keep sort close to the records list, if move to top it will be confusing to the user specially with the heading present in between sort button and list. user might think sort for the heading list

aaxelb:
@sheriefvt Okay, that's valid. In that case, there should be a little more space between "Sort" and the bottom of the heading, to look more consistent. The way it hugs the lower border looks wrong.

sheriefvt:
@aaxelb added space below sort
image

aaxelb (approved):
🇰🇮

PR #92, opened 2017-11-09
[MOD-165][Improvement] Add pending number next to moderation on dashboard

laurenbarker:
## Purpose
Display pending count next to moderation on dashboard.

## Changes
* Added request to get the pending number for each provider.
* Added skeleton screen for dashboard sidebar.

### Before
screen shot 2017-11-09 at 3 27 06 pm

### After
screen shot 2017-11-09 at 3 04 35 pm

### DIFF
--- a/app/components/provider-links/component.js
+++ b/app/components/provider-links/component.js
@@ -1,0 +1,30 @@
+import { inject as service } from '@ember/service';
+import Component from '@ember/component';
+import { task } from 'ember-concurrency';
+
+export default Component.extend({
+    store: service(),
+    theme: service(),
+
+    tagName: 'ul',
+    classNames: ['fa-ul'],
+
+    init() {
+        this._super(...arguments);
+        this.iconMap = {
+            preprint: 'fa-graduation-cap',
+            paper: 'fa-file-text-o',
+            thesis: 'fa-book',
+        };
+        this.pendingCount = null;
+        this.get('fetchData').perform();
+    },
+
+    fetchData: task(function* () {
+        const provider = yield this.get('theme').loadProvider(this.get('provider.id'));
+        const results = yield this.get('store').queryHasMany(provider, 'preprints', {
+            'meta[reviews_state_counts]': true,
+        });
+        this.set('pendingCount', results.meta.reviews_state_counts.pending);
+    }),
+});
--- a/app/components/provider-links/component.js
+++ b/app/components/provider-links/component.js
@@ -1,10 +1,10 @@
 import { inject as service } from '@ember/service';
+import { isEmpty } from '@ember/utils';
 import Component from '@ember/component';
-import { task } from 'ember-concurrency';
+import { task, timeout } from 'ember-concurrency';
 
 export default Component.extend({
     store: service(),
-    theme: service(),
 
     tagName: 'ul',
     classNames: ['fa-ul'],
@@ -21,10 +21,12 @@ export default Component.extend({
     },
 
     fetchData: task(function* () {
-        const provider = yield this.get('theme').loadProvider(this.get('provider.id'));
-        const results = yield this.get('store').queryHasMany(provider, 'preprints', {
-            'meta[reviews_state_counts]': true,
+        const results = yield this.get('store').findRecord('preprint-provider', this.get('provider.id'), {
+            adapterOptions: {
+                query: { related_counts: true },
+            },
+            reload: true,
         });
-        this.set('pendingCount', results.meta.reviews_state_counts.pending);
+        this.set('pendingCount', results.get('reviewableStatusCounts.pending'));
     }),
 });
--- a/app/components/provider-links/component.js
+++ b/app/components/provider-links/component.js
@@ -1,13 +1,11 @@
 import { inject as service } from '@ember/service';
-import { isEmpty } from '@ember/utils';
 import Component from '@ember/component';
-import { task, timeout } from 'ember-concurrency';
+import { task } from 'ember-concurrency';
 
 export default Component.extend({
     store: service(),
 
-    tagName: 'ul',
-    classNames: ['fa-ul'],
+    tagName: 'li',
 
     init() {
         this._super(...arguments);
--- a/app/components/provider-links/style.scss
+++ b/app/components/provider-links/style.scss
@@ -1,8 +1,6 @@
 & {
-  .provider-list-item {
-    margin-top: 15px;
-    padding-bottom: 10px;
-  }
+  margin-top: 15px;
+  padding-bottom: 10px;
 
   .icon {
     padding-top: 2px;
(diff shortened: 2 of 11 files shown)


app/components/provider-links/component.js
```diff
@@ -0,0 +1,30 @@
+import { inject as service } from '@ember/service';
+import Component from '@ember/component';
+import { task } from 'ember-concurrency';
+
+export default Component.extend({
+    store: service(),
+    theme: service(),
+
+    tagName: 'ul',
```
aaxelb: These are now a bunch of lists with one item each, which seems weird.

app/components/provider-links/component.js
```diff
@@ -0,0 +1,30 @@
...
+            thesis: 'fa-book',
+        };
+        this.pendingCount = null;
+        this.get('fetchData').perform();
+    },
+
+    fetchData: task(function* () {
+        const provider = yield this.get('theme').loadProvider(this.get('provider.id'));
+        const results = yield this.get('store').queryHasMany(provider, 'preprints', {
+            'meta[reviews_state_counts]': true,
```
aaxelb: The counts are [available on the provider](https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/models/preprint-provider.js#L25), if you use `?related_counts=true`.

That only works when fetching one provider at a time, which would be fine, but if we wanted we could make a [two-line change to the backend](https://github.com/CenterForOpenScience/osf.io/blob/develop/api/actions/serializers.py#L38) (would have sworn it already worked in the list view, alas) and get all the counts in that one request in `load-providers`.

app/components/provider-links/component.js
```diff
@@ -0,0 +1,30 @@
...
+            preprint: 'fa-graduation-cap',
+            paper: 'fa-file-text-o',
+            thesis: 'fa-book',
+        };
+        this.pendingCount = null;
+        this.get('fetchData').perform();
+    },
+
+    fetchData: task(function* () {
+        const provider = yield this.get('theme').loadProvider(this.get('provider.id'));
```
aaxelb: It feels wrong to use `theme` on this non-provider-specific view. Could just ask `store` for it.

aaxelb (changes requested):
Could cut the new requests in half, if not get rid of them entirely (with a backend change)

app/components/provider-links/component.js
laurenbarker: Ahhhhh I was looking for that and I never found it...

app/components/provider-links/component.js
laurenbarker: Yeah I tried using `store` first but it didn't work. I'll try again though because it might have been user error.

laurenbarker:
@aaxelb, I considered changing the backend but we probably won't be able to get that released for a long time. What about adding a ticket for the backend changes in the next sprint as an improvement to this?

aaxelb:
@laurenbarker Sounds good to me. Depending how fast it loads with one request per provider, it might not be important to optimize.

laurenbarker:
I'm not sure how to write tests for components that use ember-concurrency... I think it would involve refactoring the component so that the task is passed in. That way the task can be overwritten in tests.

app/styles/_util.scss
```diff
@@ -2,3 +2,11 @@
   pointer-events: none;
   opacity: 0.4;
 }
+
+.unread-label {
+    background-color: $color-bg-gray-light;
```
aaxelb: Are we back to 4-space indents for CSS?

app/components/provider-links/style.scss
```diff
@@ -0,0 +1,42 @@
+& {
```
aaxelb, line 1: Doesn't ember-component-css take care of making sure these styles only apply to this component? Do we need to wrap everything in `& {}`?

aaxelb (changes requested):
Tiny nitpicks, otherwise 🇫🇯

app/styles/_util.scss
laurenbarker: Oops... We should add a linter for that.

app/components/provider-links/style.scss
laurenbarker: It's not for scoping; it allows you to set css on component. So really the only thing that needs to go in `& {}` is the styling for the actual component. Should I pull everything else out?

aaxelb: Ah, right, I skimmed right over the styles and just saw the classes. I guess it seems cleaner to use `&` only for the styles on the component element itself, and pull everything else out, but that's not a strong preference.

laurenbarker: @aaxelb, they have been un-nested.

app/components/provider-links/template.hbs
```diff
@@ -0,0 +1,39 @@
...
+    <span class="provider-name {{if provider.reviewsWorkflow "moderation-enabled" "moderation-disabled"}}">
+        <i class="fa-li fa {{get iconMap provider.preprintWord}} icon"></i>
+        {{provider.name}}
+    </span>
+    <ul class="provider-links">
+        {{#if provider.reviewsWorkflow}}
+            <li>
+                {{#link-to 'preprints.provider.moderation' provider.id}}
+                    {{t 'global.moderation'}}
+                {{/link-to}}
```
aaxelb: FYI, because you took the time to remove the underline on the moderation count, you could get rid of this extended underline:
screen shot 2017-11-14 at 3 12 55 pm
by stripping the whitespace inside the `link-to` with [strategic use of tildes](http://handlebarsjs.com/expressions.html#whitespace-control)

app/components/provider-links/template.hbs
```diff
@@ -0,0 +1,39 @@
+{{#if fetchData.isRunning}}
+    {{#content-placeholders as |placeholder|}}
+        <span>
+            <i class="fa-li icon--skeleton">{{placeholder.icon}}</i>
+            {{placeholder.heading subtitle=false}}
```
aaxelb, line 5: It looks like it's still adding a subtitle, and it looks weird:
screen shot 2017-11-14 at 3 11 54 pm

aaxelb (changes requested):
I lied, I found a couple more nits to pick. Otherwise, 🇻🇺

aaxelb:
Oh, and add this to the changelog, please!

app/components/provider-links/template.hbs
laurenbarker: Did you run `yarn install`? I made a change to `ember-content-placeholders`.

aaxelb: Oh whoops, I did not. All good 👍

aaxelb (approved):
🇹🇻

PR #94, opened 2017-11-15
[MOD-186][Improvement] Add skeleton for moderation-list-row

laurenbarker:
## Purpose
Improve loading on the moderation-list page.

## Changes
* Add loading page for provider route
    * I couldn't figure out a way around loading the theme in the provider route.
* Add `moderation-list-row-skeleton`
* Install ember-parachute
screen shot 2017-11-15 at 5 42 25 pm


app/components/moderation-list-row/template.hbs
```diff
@@ -1,12 +1,11 @@
-{{#link-to 'preprints.provider.preprint-detail' theme.id submission.id
-           disabled=(or moderatorLoading submission.node.isPending submission.node.contributors.isPending)}}
-    <div data-status={{submission.reviewsState}} class="flex-container">
-        <i class="fa fa-lg {{get iconClass submission.reviewsState}} status-icon"></i>
-        <div title={{submission.title}} class="submission-info">
-            <div class="submission-title">{{submission.title}}</div>
-            {{#if (or moderatorLoading submission.node.isPending submission.node.contributors.isPending)}}
-                {{loading-indicator classes='ball-dark'}}
-            {{else}}
+{{#if (or moderatorLoading submission.node.isPending submission.node.contributors.isPending)}}
```
aaxelb: Is there a way to wrap all the data loading in a task instead of all these `isPending`s? This line makes me sad.

app/components/moderation-list/template.hbs
```diff
@@ -29,7 +29,7 @@
                     {{/bs-dropdown}}
                 </div>
                 {{#if loading}}
-                    {{loading-indicator classes='ball-dark text-center m-b-lg p-v-md'}}
+                    {{moderation-list-row-skeleton}}
                 {{else}}
                     <div class="moderation-list-body">
```
aaxelb: If the screen is short (or the list is long), the margin jumps around during loading:
kapture 2017-11-20 at 15 36 03

app/controllers/preprints/provider/moderation.js
```diff
@@ -1,26 +1,41 @@
...
+        defaultValue: 'pending',
+        refresh: true,
+    },
+});
+
+export default Controller.extend(Analytics, moderationQueryParams.Mixin, {
+    store: service(),
+    theme: service(),
+
+    queryParamsChanged: computed.or('queryParamsState.{page,sort,status}.changed'),
```
aaxelb: Is this necessary? It doesn't look used.

app/routes/preprints/provider.js
```diff
@@ -13,6 +13,7 @@ export default Route.extend({
     theme: service(),
 
     model(params) {
+        // need to load the provider here so theme can be used in child routes
         return this.get('theme').loadProvider(params.provider_id)
```
aaxelb, line 17: I'm okay with loading the provider in the route, but could we do it without using `theme`? And pass the provider into the components that need it?

A nice thing about that is we could link to this page from the dashboard without the route hooks running, since we already have the provider with counts and everything.

app/components/moderation-base/component.js
```diff
@@ -1,30 +1,57 @@
...
+                count: this.get('pendingCount'),
             },
             {
                 nameKey: 'global.settings',
                 route: 'preprints.provider.settings',
             },
         ];
     }),
+
+    init() {
```
aaxelb: This might make more sense as `didReceiveAttrs`. `init` makes tests harder and shouldn't be used for stuff that depends on incoming values.

aaxelb (changes requested):
Overall looks good! Just a few thoughts.

app/components/moderation-base/component.js
laurenbarker: I moved it from a controller and forgot to change it.

app/components/moderation-list/template.hbs
laurenbarker: Good catch. That only happens when the screen is really short.

app/controllers/preprints/provider/moderation.js
laurenbarker: IT'S NOT! I feel betrayed by the `ember-parachute` example...

app/routes/preprints/provider.js
laurenbarker: The issue with the provider is that it gets overwritten in the store all the time. So if there is a request to the `preprint_provider` endpoint that doesn't get the review counts then all of a sudden the counts aren't there.

aaxelb: If it's consistently loaded in the route and/or dashboard (both with counts), and we're careful not to reload it otherwise, won't the store continue to reuse the same provider (with counts)?

...I guess even if that's true, it's fragile and requires that we remember to give providers special treatment. What if we extend the provider adapter to add `meta[reviews_state_counts]=true` to all requests, so we don't need to think about it elsewhere?

laurenbarker: The counts are in different locations too. The request on the dashboard page and the request on the provider page overwrite each other in the store. I think it would require backend changes (although probably a good idea). Below are the provider requests.

```js
const results = yield this.get('store').findRecord('preprint-provider', this.get('provider.id'), {
    adapterOptions: {
        query: { related_counts: true },
    },
    reload: true,
});
this.set('pendingCount', results.get('reviewableStatusCounts.pending'));
```

I don't think there is way to get the counts here?
```js
const results = yield this.get('store').query('preprint-provider', query);
return this.set('providers', results);
```

The request for the provider here could probably be updated to include the counts.
```js
const provider = yield this.get('store').findRecord(
    'preprint-provider',
    this.get('theme.id'),
);
const response = yield this.get('store').queryHasMany(provider, 'preprints', {
    filter: {
        reviews_state: queryParams.status,
        node_is_public: true,
    },
    'meta[reviews_state_counts]': true,
    sort: queryParams.sort,
    page: queryParams.page,
});
this.set('results', {
    submissions: response.toArray(),
    totalPages: Math.ceil(response.meta.total / response.meta.per_page),
    statusCounts: response.meta.reviews_state_counts,
});
```

app/components/moderation-list-row/component.js
```diff
@@ -73,13 +72,18 @@ export default Component.extend({
...
             pending: 'fa-hourglass-o pending',
             rejected: 'fa-times-circle-o rejected',
         };
+        this.get('fetchData').perform();
     },
+
+    fetchData: task(function* () {
+        const node = yield this.get('submission.node');
+        yield node.get('contributors');
+        yield this.get('submission.actions');
```
aaxelb, line 85: ✨

app/components/moderation-base/component.js
```diff
@@ -1,30 +1,56 @@
...
+            const response = yield this.get('store').findRecord(
+                'preprint-provider',
+                this.get('theme.id'), {
+                    adapterOptions: {
+                        query: { related_counts: true },
+                    },
+                    reload: true,
+                },
+            );
+            this.set('pendingCount', response.get('reviewableStatusCounts.pending'));
```
aaxelb: `pendingCount` needs to be updated when we get fresh counts from a moderation list page. If a preprint is submitted while this page is open, the counts in the table will be updated, but the tab above won't until the page is refreshed:
kapture 2017-11-21 at 14 50 08

Maybe use an action?

app/components/moderation-list-row/component.js
```diff
@@ -34,16 +37,12 @@ export default Component.extend({
...
-    }),
-
     latestAction: computed('submission.actions.[]', function() {
         return latestAction(this.get('submission.actions'));
     }),
 
-    noActions: computed('submission.actions.{length,isPending}', function () {
-        return !this.get('submission.actions.length') || this.get('submission.actions.isPending');
+    noActions: computed('submission.actions.length', function () {
+        return !this.get('submission.actions.length');
```
aaxelb: `computed.not`?

app/routes/preprints/provider.js
aaxelb: How would you feel about loading the provider in a controller task (triggered by this route's `setupController`) and passing the provider down to the components that need it?

app/components/moderation-base/component.js
```diff
@@ -1,30 +1,56 @@
...
             {
                 nameKey: 'global.settings',
                 route: 'preprints.provider.settings',
             },
         ];
     }),
+
+    didReceiveAttrs() {
+        this.pendingCount = this.get('theme.provider.reviewableStatusCounts.pending');
+        this.providerName = this.get('theme.provider.name');
```
aaxelb: Should these use `set`?

app/routes/preprints/provider.js
laurenbarker: If `theme.provider` is loaded in the provider controller then the provider can be accessed through the theme and doesn't need to be passed down. That way it would be consistent with `theme.reviewableStatusCounts`.

aaxelb: Makes sense 👍

laurenbarker: So I remembered why this doesn't work. The dynamic path is only passed to the model hook so even if setupController is used to call a task on the controller, the provider id isn't going to be available.

aaxelb: Oh, and setupController gets the result of the model hook. I suppose `model` could just return the parameter that setupController needs, and call that the model, but that might be ugly.

app/components/moderation-base/component.js
```diff
@@ -1,30 +1,58 @@
...
     }),
+
+    didReceiveAttrs() {
+        this.set('providerName', this.get('theme.provider.name'));
+        this.get('fetchData').perform();
+    },
+
+    fetchData: task(function* () {
+        try {
+            const response = yield this.get('store').findRecord(
```
aaxelb: This looks identical to calling `theme.loadProvider`, could we do that instead?

...but also, do we need to refresh the counts here? I think it'll usually run right after the route model hook refreshes the provider.

laurenbarker: Oh probably. I'll test it out

laurenbarker: @aaxelb, done 💰

app/components/moderation-base/component.js
```diff
@@ -1,30 +1,36 @@
...
             },
             {
                 nameKey: 'global.settings',
                 route: 'preprints.provider.settings',
             },
         ];
     }),
+
+    didReceiveAttrs() {
+        this.set('providerName', this.get('theme.provider.name'));
```
aaxelb: This would make more sense as a computed alias, since `theme.provider` updating wouldn't rerun `didReceiveAttrs`

aaxelb (dismissed):
Just the tiniest nitpick (I've learned I like aliases), but feel free to ignore and merge if you'd rather

laurenbarker:
@aaxelb, updated 👍

aaxelb (approved):

PR #98, opened 2017-11-28
[MOD-187][Improvement] Preprint Detail Loading

laurenbarker:
## Purpose
Handle loading properly on moderation detail page.

## Changes
Biggest change is how contributors are loaded on the detail page. There are 3 scenarios:
1. transition from another page via link
    * contributors are not embedded must load all of them
    * user will see skeleton screens
2. refresh current page < 10 contributors
    * contributors are embedded don't need to load more
    * user will not see skeleton screen
3. refresh current page > 10 contributors
    * first 10 are embedded but need to get all of them
    * user will see skeleton screen

*Note that you won't see the loading indicator if there are less than 10 contributors (unlike this gif). Also, the loading indicator is replaced with a skeleton screen.*

preprint-detail-loading


app/components/contributor-list/style.scss
```diff
@@ -1,20 +1,28 @@
 & {
+  // color: $color-white;
```
aaxelb: Delete?

app/components/contributor-list/component.js
```diff
@@ -1,21 +1,46 @@
...
+        const node = this.get('node.content');
+        const query = {
+            'page[size]': 100,
+            page: this.get('page'),
+        };
+
+        let response = yield this.get('loadContributors').perform(node, query);
+
+        while (response.links.next) {
+            query.page = this.incrementProperty('page');
```
aaxelb: It seems like `page` would make more sense as a local variable (or just `query.page++`). Is there a reason it's kept as a property on the component? I think we should generally limit side effects of tasks.

app/controllers/preprints/provider/preprint-detail.js
```diff
@@ -133,4 +140,26 @@ export default Controller.extend({
...
     },
+
+    fetchData: task(function* (preprintId) {
+        const response = yield this.get('store').findRecord(
+            'preprint',
+            preprintId,
+            { include: ['node', 'license', 'actions', 'contributors'] },
+        );
+        if (!response.get('node.public')) {
+            this.transitionTo('page-not-found');
```
aaxelb: 1. `transitionToRoute` (or `router` service?)
2. When transitioning from the dashboard, the node isn't present and this is erroneously hit
3. Might not matter, but should probably return after triggering a transition

app/components/contributor-list/component.js
```diff
@@ -1,21 +1,46 @@
...
     node: null,
+    page: 1,
 
     bibliographicContributors: filterBy('contributors', 'bibliographic', true),
 
-    contributors: computed('node', function() {
-        return loadRelation(this.get('node'), 'contributors');
+    init() {
+        this._super(...arguments);
+        this.set('contributorsList', this.get('contributors') || []);
```
aaxelb: Stuff that depends on passed-in attributes should be in `didReceiveAttrs`, not `init`

app/components/contributor-list/component.js
```diff
@@ -1,21 +1,46 @@
...
+        if (!this.get('contributors.length') || this.get('contributors.content.meta.total') > this.get('contributors.content.meta.per_page')) {
```
aaxelb: Would it make more sense to check `contributors.content.meta.total > contributors.length`, in case all the contributors are already in the store?

app/components/contributor-list/component.js
```diff
@@ -1,21 +1,46 @@
 import { filterBy } from '@ember/object/computed';
-import { computed } from '@ember/object';
 import Component from '@ember/component';
-import { loadRelation } from 'ember-osf/utils/load-relationship';
```
aaxelb, line 4: The only remaining use of `loadRelation` is in `preprint-file-browser`, should we remove that too?

aaxelb (changes requested):
Looks pretty good! And such a stylish PR. Just had a few thoughts

app/components/contributor-list/component.js
laurenbarker: Should probably remove it during [MOD-176](https://openscience.atlassian.net/browse/MOD-176) that moves queryHasMany to ember-osf.

aaxelb (commented):
Looks good to me, but could you add ~2% worth of tests (maybe just merging develop would do that) and resolve conflicts?

aaxelb (approved):
💀

PR #103, opened 2017-12-11
[IN-16][Reviews] Update frontend to reflect divorce changes

sheriefvt:
## Goal
Update preprint requests using the node endpoints to use the preprint endpoints.
This is the first step of the preprint node divorce.

## Ticket
https://openscience.atlassian.net/browse/IN-16

### DIFF
--- a/app/components/contributor-list/component.js
+++ b/app/components/contributor-list/component.js
@@ -1,21 +1,8 @@
 import { filterBy } from '@ember/object/computed';
-import { computed } from '@ember/object';
 import Component from '@ember/component';
-import { loadRelation } from 'ember-osf/utils/load-relationship';
 
-const ContributorListComponent = Component.extend({
+export default Component.extend({
     tagName: 'ul',
-    node: null,
 
-    bibliographicContributors: filterBy('contributors', 'bibliographic', true),
-
-    contributors: computed('node', function() {
-        return loadRelation(this.get('node'), 'contributors');
-    }),
+    bibliographicContributors: filterBy('model.contributors', 'bibliographic', true),
 });
-
-ContributorListComponent.reopenClass({
-    positionalParams: ['node'],
-});
-
-export default ContributorListComponent;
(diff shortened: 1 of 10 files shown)


laurenbarker:
@sheriefvt, just as a heads up, I think you're going to have merge conflicts with https://github.com/CenterForOpenScience/ember-osf-reviews/pull/98.

sheriefvt:
@laurenbarker 👍 thanks

app/templates/preprints/provider/preprint-detail.hbs
```diff
@@ -5,7 +5,7 @@
                 <div class="col-xs-12 col-sm-11">
                     {{reviews-breadcrumbs}}
                     <h1 id="preprintTitle" class="p-t-md p-b-md">{{node.title}}</h1>
-                    <h5>{{contributor-list node}}</h5>
+                    <h5>{{contributor-list model=model}}</h5>
```
aaxelb, line 8: Generally speaking, we shouldn't use the name `model` outside of a route template. Choose a more descriptive name, like maybe `preprint` or `submission`

tests/unit/components/moderation-list-row-test.js
```diff
@@ -0,0 +1,96 @@
...
+import { moduleForComponent } from 'ember-qunit';
+import test from 'ember-sinon-qunit/test-support/test';
+import { freezeDateAt, unfreezeDate } from 'ember-mockdate-shim';
+
+moduleForComponent('moderation-list-row', 'Unit | Component | moderation list row', {
+
+    unit: true,
+    needs: [
+        'model:action',
+        'model:node',
```
aaxelb, line 11: Does this actually need `model:node`? It seems like the point of the PR is to get rid of that dependency.

app/components/contributor-list/component.js
```diff
@@ -1,21 +1,8 @@
...
 
-const ContributorListComponent = Component.extend({
+export default Component.extend({
     tagName: 'ul',
-    node: null,
 
-    bibliographicContributors: filterBy('contributors', 'bibliographic', true),
-
-    contributors: computed('node', function() {
-        return loadRelation(this.get('node'), 'contributors');
```
aaxelb, line 13: Don't worry about fixing this, because @laurenbarker rewrote this component in #98, but note that the old code would load every page of contributors using `loadRelation`. Changing it to use `model.contributors` would just show the first page.

aaxelb (commented):
It looks like the `preprint-detail` route/controller/template and the `preprint-file-browser` component still use `preprint.node`, could you update those too?

alexschiller:
Closing this PR-- #126 branched off and updated it.

PR #104, opened 2017-12-14
[Feature][MOD-181] Give warning on moderation detail page if reviewer input not submitted

sheriefvt:
## Problem
If a reviewer accidentally exits the decision form we don't want them to navigate away from the page without making a decision.

## Goal
Give warning on moderation detail page if reviewer input (e.g. new/edit comment or change of decision) not submitted and reviewer tries to leave the page.

## Changes

- Update preprint-status-banner component
- Update preprint-detail page template, controller, and router
- Update translation file
- Add new addon ember-onbeforeunload for case 1 (see next)
- Add new unit tests

**Two cases**
- When user reloads or closes the page or navigate to OSF or another external app

- When user navigates to another route with-in the reviews app

## Ticket 
https://openscience.atlassian.net/browse/MOD-181

### DIFF
--- a/app/components/preprint-status-banner/component.js
+++ b/app/components/preprint-status-banner/component.js
@@ -262,11 +262,7 @@ export default Component.extend({
             this.set('reviewerComment', this.get('initialReviewerComment'));
         },
         closeReviewerFeedback() {
-            if (this.get('decisionChanged') || this.get('commentEdited')){
-                this.set('userHasEnteredReview', true);
-            } else {
-                this.set('userHasEnteredReview', false);
-            }
+            this.set('userHasEnteredReview', this.get('userActivity'));
         },
     },
 
--- a/app/components/preprint-status-banner/component.js
+++ b/app/components/preprint-status-banner/component.js
@@ -231,10 +231,7 @@ export default Component.extend({
     }),
 
     btnDisabled: computed('decisionChanged', 'commentEdited', 'saving', 'commentExceedsLimit', function() {
-        if (this.get('saving') || (!this.get('decisionChanged') && !this.get('commentEdited')) || this.get('commentExceedsLimit')) {
-            return true;
-        }
-        return false;
+        return this.get('saving') || (!this.get('decisionChanged') && !this.get('commentEdited')) || this.get('commentExceedsLimit');
     }),
 
     didInsertElement() {
--- a/app/components/preprint-status-banner/component.js
+++ b/app/components/preprint-status-banner/component.js
@@ -121,6 +121,8 @@ export default Component.extend({
 
     commentExceedsLimit: computed.gt('reviewerComment.length', COMMENT_LIMIT),
 
+    userActivity: computed.or('commentEdited', 'decisionChanged'),
+
     commentLengthErrorMessage: computed('reviewerComment', function () {
         const i18n = this.get('i18n');
         return i18n.t('components.preprintStatusBanner.decision.commentLengthError', {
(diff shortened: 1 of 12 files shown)


app/templates/preprints/provider/preprint-detail.hbs
```diff
@@ -136,4 +136,17 @@
 
         </div>
     </div>
+
+    {{#if showWarning}}
+        {{#bs-modal onHide=(action 'hideWarning') as |modal|}}
```
aaxelb: I think if you used `open=showWarning` instead of `{{#if showWarning}}`, bootstrap could do a nicer CSS transition.

app/components/preprint-status-banner/component.js
```diff
@@ -227,6 +227,10 @@ export default Component.extend({
         return this.get('submission.reviewsState') !== this.get('decision');
     }),
 
+    userActivity: computed('commentEdited', 'decisionChanged', function () {
```
aaxelb: This could be `computed.or`.

app/templates/preprints/provider/preprint-detail.hbs
```diff
@@ -136,4 +136,17 @@
...
+            {{#modal.header}}
+                <div class="warning-header"><i class="fa fa-exclamation-triangle"></i> {{t 'content.warning.header'}}</div>
+            {{/modal.header}}
+            {{#modal.body}}<div class="warning-body">{{t 'content.warning.body'}}</div>{{/modal.body}}
+            {{#modal.footer}}
+                {{#bs-button onClick=(action 'leavePage')}}{{t 'content.warning.footer.leave'}}{{/bs-button}}
+                {{#bs-button onClick=(action 'hideWarning') type="info"}}{{t 'content.warning.footer.stay'}}{{/bs-button}}
```
aaxelb: These buttons are switched compared to the buttons in the browser modal:
screen shot 2017-12-15 at 2 03 54 pm
screen shot 2017-12-15 at 2 03 39 pm
screen shot 2017-12-15 at 2 29 59 pm
screen shot 2017-12-15 at 2 30 10 pm

app/components/preprint-status-banner/component.js
```diff
@@ -257,6 +261,9 @@ export default Component.extend({
             this.set('decision', this.get('submission.reviewsState'));
             this.set('reviewerComment', this.get('initialReviewerComment'));
         },
+        closeReviewerFeedback() {
+            this.set('userHasEnteredReview', this.get('userActivity'));
```
aaxelb, line 260: Because of how this is determined, and because "Accept" is selected by default, there might be a minor bug: If you go to a pending submission, then open and close the dropdown without changing anything, `userHasEnteredReview` will be true.

app/styles/modules/provider/_preprint-detail.scss
```diff
@@ -12,3 +12,16 @@ section.preprint-license {
     width: 100%;
   }
 }
+
+.warning-header {
+  font-size: 20px;
+  padding: 5px 15px 8px 15px;
+
+  a {
+    color: grey;
```
aaxelb: Use variables for color

aaxelb (changes requested):
Overall, looks good! Had a few suggestions and maybe found a small bug.

app/templates/preprints/provider/preprint-detail.hbs
sheriefvt: In the ticket --> developer notes, the "leave page" is on the left. I agree the order should be consistent with the browser modal, will update.

app/components/preprint-status-banner/component.js
```diff
@@ -224,14 +226,12 @@ export default Component.extend({
     }),
 
     decisionChanged: computed('submission.reviewsState', 'decision', function() {
-        return this.get('submission.reviewsState') !== this.get('decision');
+        return ((this.get('submission.reviewsState') !== this.get('decision') && this.get('submission.reviewsState') !== PENDING) ||
+        (this.get('submission.reviewsState') === PENDING && this.get('decision') !== ACCEPTED));
```
aaxelb: Stop me if I'm getting too picky, but this is getting hard to read, and I'd like to avoid state-dependent logic when we can -- it might be cleaner to save an `initialDecision` somewhere (similar to `initialReviewerComment`), and check `decision !== initialDecision`

app/routes/preprints/provider/preprint-detail.js
```diff
@@ -45,6 +46,19 @@ export default Route.extend({
                 return this.intermediateTransitionTo('page-not-found');
             }
         },
+        willTransition(transition) {
+            if (this.controller.get('userHasEnteredReview')) {
```
aaxelb: Could this reuse `this.isPageDirty()`? It could someday get more complex, it would be nice to have it in a single place.

sheriefvt:
Replace this PR with #108 which is based on future-reviews/0.4.0 branch

PR #105, opened 2017-12-15
[MOD-197][Improvement] actions --> reviewActions

laurenbarker:
*The ember-osf-reviews changes rely on the API changes and ember-osf changes in the related tickets below.*

## Purpose
Change `actions` --> `reviewActions` to support generalizing the actions API.

Related tickets: 
* https://github.com/CenterForOpenScience/osf.io/pull/7924 but with updated changes https://github.com/mfraezz/osf.io/tree/feature/request-actions
* https://github.com/CenterForOpenScience/ember-osf/pull/329
    * will need to be merged into our version of ember-osf (they aren't merged quite yet)

## Changes
* mainly just update the name
* actions for the actions-list are loaded from `reviews/actions` instead of the user relationship

## QA notes
* regression for moderation functionality; everything should work the same

### DIFF
--- a/app/components/action-feed/component.js
+++ b/app/components/action-feed/component.js
@@ -1,7 +1,6 @@
 import { computed } from '@ember/object';
 import { inject as service } from '@ember/service';
 import Component from '@ember/component';
-import { translationMacro as t } from 'ember-i18n';
 import { task } from 'ember-concurrency';
 
 /**
@@ -17,11 +16,11 @@ import { task } from 'ember-concurrency';
 export default Component.extend({
     store: service(),
     toast: service(),
+    i18n: service(),
+
     currentUser: service(),
     classNames: ['action-feed'],
     page: 0,
-
-    errorMessage: t('components.actionFeed.errorLoading'),
 
     dummyActionList: computed(function() {
         return new Array(10);
@@ -42,7 +41,7 @@ export default Component.extend({
         const page = this.get('page');
         try {
             const user = yield this.get('currentUser.user');
-            const actions = yield this.get('store').queryHasMany(user, 'actions', {
+            const actions = yield this.get('store').queryHasMany(user, 'reviewActions', {
                 page,
                 embed: 'target',
             });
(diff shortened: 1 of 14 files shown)


aaxelb:
@laurenbarker Would you mind setting the ember-osf dependency to that PR's branch, so travis can actually run? We'll have to update the dependency anyway once those changes are released.

app/components/action-feed/component.js
```diff
@@ -41,18 +40,18 @@ export default Component.extend({
...
+            const actions = yield this.get('store').query('review-action', {
                 page,
-                embed: 'target',
+                embed: ['target'],
             });
             this.get('actionsList').pushObjects(actions.toArray());
             this.set(
                 'totalPages',
-                Math.ceil(actions.get('meta.total') / actions.get('meta.per_page')),
+                Math.ceil(actions.get('meta.pagination.total') / actions.get('meta.pagination.per_page')),
```
aaxelb, line 50: Which version of the API uses `meta.pagination`? Is that new?

aaxelb (commented):
Looks good, just one question

app/components/action-feed/component.js
laurenbarker: That is a great question. When I was testing it out `meta.total` was incorrect but `meta.pagination.total` was an accurate number. I'll look into why it showed up.

laurenbarker: https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/serializers/osf-serializer.js#L143 (h/t @jamescdavis)

jamescdavis: `meta.total` is actually total pages. It should probably be renamed to `total_pages` so this is clear.

aaxelb: Ok, I think this is fine for now. If/when `total_pages` is available, we should use that instead.

aaxelb (approved):

PR #107, opened 2017-12-21
[Feature][MOD-191] Add submitted by to the accepted/rejected records

sheriefvt:
## Goal
In the rejected/accepted section on the moderation list view, add submitted by along with the accepted by/rejected by.
"submission by <CONTRIBUTORS (3)> was <accepted|rejected> on <DATE> by <MODERATOR>"

WHY:
In demo of accepted/rejected by we discussed recent feedback from users that requested the submitted by author name should also be there.

## Changes

- Update 'moderation list row' component, css, and related translations
- Update component integration and unit tests

image

image

## Ticket
https://openscience.atlassian.net/browse/MOD-191

### DIFF
--- a/app/components/moderation-list-row/component.js
+++ b/app/components/moderation-list-row/component.js
@@ -54,29 +54,22 @@ export default Component.extend({
         return this.get('submission.node.contributors.content.meta.pagination.total') - 3;
     }),
 
-    gtDay: computed('submission.dateLastTransitioned', function() {
-        return moment().diff(this.get('submission.dateLastTransitioned'), 'days') > 1;
-    }),
-
-    // date of preprint acceptance or rejection
-    acceptedRejectedDate: computed('submission.dateLastTransitioned', 'gtDay', function() {
-        return this.get('gtDay') ?
+    // translations for moderator action label
+    reviewedOnLabel: computed('submission.{reviewsState,dateLastTransitioned}', 'noActions', 'latestActionCreator', function() {
+        const gtDay = moment().diff(this.get('submission.dateLastTransitioned'), 'days') > 1;
+        const acceptedRejectedDate = gtDay ?
             moment(this.get('submission.dateLastTransitioned')).format('MMMM DD, YYYY') :
             moment(Math.min(Date.parse(this.get('submission.dateLastTransitioned')), Date.now())).fromNow();
-    }),
-
-    // translations for moderator action label
-    reviewedOnLabel: computed('submission.reviewsState', 'gtDay', 'noActions', 'acceptedRejectedDate', 'latestActionCreator', function() {
         const i18n = this.get('i18n');
-        const dayValue = this.get('gtDay') ? 'gtDay' : 'ltDay';
+        const dayValue = gtDay ? 'gtDay' : 'ltDay';
         const timeWording = this.get('noActions') ? `${dayValue}_automatic` : dayValue;
         const status = this.get('submission.reviewsState');
         const labels = ACTION_LABELS[status][timeWording];
-        return i18n.t(labels, { timeDate: this.get('acceptedRejectedDate'), moderatorName: this.get('latestActionCreator') });
+        return i18n.t(labels, { timeDate: acceptedRejectedDate, moderatorName: this.get('latestActionCreator') });
     }),
 
     // translations for submitted on label
-    submittedOnLabel: computed('gtDay', 'submission.dateCreated', function() {
+    submittedOnLabel: computed('submission.dateCreated', function() {
         const i18n = this.get('i18n');
         const gtDaySubmit = moment().diff(this.get('submission.dateCreated'), 'days') > 1;
         const dayValue = gtDaySubmit ? 'gtDay' : 'ltDay';
--- a/app/components/moderation-list-row/component.js
+++ b/app/components/moderation-list-row/component.js
@@ -37,11 +37,11 @@ export default Component.extend({
 
     latestActionCreator: computed.alias('latestAction.creator.fullName'),
 
-    noActions: computed.not('submission.actions.length'),
+    noActions: computed.not('submission.reviewActions.length'),
 
     // latest action attempted on the preprint
-    latestAction: computed('submission.actions.[]', function() {
-        return latestAction(this.get('submission.actions'));
+    latestAction: computed('submission.reviewActions.[]', function() {
+        return latestAction(this.get('submission.reviewActions'));
     }),
 
     // first three contributors to the preprint
@@ -54,41 +54,38 @@ export default Component.extend({
         return this.get('submission.node.contributors.content.meta.pagination.total') - 3;
     }),
 
-    // check if it has been more than a day since preprint last transition
     gtDay: computed('submission.dateLastTransitioned', function() {
         return moment().diff(this.get('submission.dateLastTransitioned'), 'days') > 1;
     }),
 
     // date of preprint acceptance or rejection
-    acceptedRejectedtDate: computed('submission.dateLastTransitioned', 'gtDay', function() {
+    acceptedRejectedDate: computed('submission.dateLastTransitioned', 'gtDay', function() {
         return this.get('gtDay') ?
             moment(this.get('submission.dateLastTransitioned')).format('MMMM DD, YYYY') :
             moment(Math.min(Date.parse(this.get('submission.dateLastTransitioned')), Date.now())).fromNow();
     }),
 
-    // date of preprint submission
-    submitDate: computed('submission.dateCreated', 'gtDay', function() {
-        return this.get('gtDay') ?
-            moment(this.get('submission.dateCreated')).format('MMMM DD, YYYY') :
-            moment(Math.min(Date.parse(this.get('submission.dateCreated')), Date.now())).fromNow();
-    }),
-
-    // translations for accepted and rejected records
-    acceptedRejectedStatusTimeDate: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
+    // translations for moderator action label
+    reviewedOnLabel: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
         const i18n = this.get('i18n');
         const dayValue = this.get('gtDay') ? 'gtDay' : 'ltDay';
         const timeWording = this.get('noActions') ? `${dayValue}_automatic` : dayValue;
+        console.log(dayValue);
         const status = this.get('submission.reviewsState');
         const labels = ACTION_LABELS[status][timeWording];
-        return i18n.t(labels, { timeDate: this.get('acceptedRejectedtDate'), moderatorName: this.get('latestActionCreator') });
+        return i18n.t(labels, { timeDate: this.get('acceptedRejectedDate'), moderatorName: this.get('latestActionCreator') });
     }),
 
-    // translations for pending records
-    pendingStatusTimeDate: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
+    // translations for submitted on label
+    submittedOnLabel: computed('gtDaySubmit', 'submission.dateCreated', function() {
         const i18n = this.get('i18n');
-        const dayValue = this.get('gtDay') ? 'gtDay' : 'ltDay';
+        const gtDaySubmit = moment().diff(this.get('submission.dateCreated'), 'days') > 1;
+        const dayValue = gtDaySubmit ? 'gtDay' : 'ltDay';
+        const submitDate = gtDaySubmit ?
+            moment(this.get('submission.dateCreated')).format('MMMM DD, YYYY') :
+            moment(Math.min(Date.parse(this.get('submission.dateCreated')), Date.now())).fromNow();
         const labels = ACTION_LABELS.pending[dayValue];
-        return i18n.t(labels, { timeDate: this.get('submitDate') });
+        return i18n.t(labels, { timeDate: submitDate });
     }),
 
     didReceiveAttrs() {
@@ -103,6 +100,6 @@ export default Component.extend({
     fetchData: task(function* () {
         const node = yield this.get('submission.node');
         yield node.get('contributors');
-        yield this.get('submission.actions');
+        yield this.get('submission.reviewActions');
     }),
 });
--- a/app/components/moderation-list-row/component.js
+++ b/app/components/moderation-list-row/component.js
@@ -39,35 +39,56 @@ export default Component.extend({
 
     noActions: computed.not('submission.actions.length'),
 
+    // latest action attempted on the preprint
     latestAction: computed('submission.actions.[]', function() {
         return latestAction(this.get('submission.actions'));
     }),
 
+    // first three contributors to the preprint
     firstContributors: computed('submission.node.contributors', function() {
         return this.get('submission.node.contributors').slice(0, 3);
     }),
 
+    // count of contributors in-addition to the first three
     additionalContributors: computed('submission.node.contributors', function() {
         return this.get('submission.node.contributors.content.meta.pagination.total') - 3;
     }),
 
+    // check if it has been more than a day since preprint last transition
     gtDay: computed('submission.dateLastTransitioned', function() {
         return moment().diff(this.get('submission.dateLastTransitioned'), 'days') > 1;
     }),
 
-    relevantDate: computed('submission.dateLastTransitioned', 'gtDay', function() {
+    // date of preprint acceptance or rejection
+    acceptedRejectedtDate: computed('submission.dateLastTransitioned', 'gtDay', function() {
         return this.get('gtDay') ?
             moment(this.get('submission.dateLastTransitioned')).format('MMMM DD, YYYY') :
             moment(Math.min(Date.parse(this.get('submission.dateLastTransitioned')), Date.now())).fromNow();
     }),
 
-    // translations
-    statusTimeDate: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
+    // date of preprint submission
+    submitDate: computed('submission.dateCreated', 'gtDay', function() {
+        return this.get('gtDay') ?
+            moment(this.get('submission.dateCreated')).format('MMMM DD, YYYY') :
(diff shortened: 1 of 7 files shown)


app/components/moderation-list-row/component.js
```diff
@@ -39,35 +39,56 @@ export default Component.extend({
...
         const timeWording = this.get('noActions') ? `${dayValue}_automatic` : dayValue;
-        const labels = ACTION_LABELS[this.get('submission.reviewsState')][timeWording];
-        return i18n.t(labels, { timeDate: this.get('relevantDate'), moderatorName: this.get('latestActionCreator') });
+        const status = this.get('submission.reviewsState');
+        const labels = ACTION_LABELS[status][timeWording];
+        return i18n.t(labels, { timeDate: this.get('acceptedRejectedtDate'), moderatorName: this.get('latestActionCreator') });
+    }),
+
+    // translations for pending records
+    pendingStatusTimeDate: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
```
aaxelb: It looks like this is displayed all the time, not just for pending submissions, so the name and comment are misleading.

This is why names should describe what things are or what they do, not the application state when we expect to use them: Our expectations are often wrong or incomplete. When we use something in a new way, we shouldn't have to rename it to avoid confusion.

This is a description of when/how the record was submitted, so maybe something like `submittedLabel` or `submittedOnLabel` or `submittedOnMessage`?

app/components/moderation-list-row/component.js
```diff
@@ -39,35 +39,56 @@ export default Component.extend({
...
+        const i18n = this.get('i18n');
+        const dayValue = this.get('gtDay') ? 'gtDay' : 'ltDay';
+        const labels = ACTION_LABELS.pending[dayValue];
```
aaxelb: This `.pending` should probably be renamed too.

app/components/moderation-list-row/style.scss
```diff
@@ -42,7 +42,8 @@
     content: ", "
   }
   li:last-child:after {
-    content: "";
+    content: "\a";
+    white-space: pre;
```
aaxelb: This is a neat trick to force a newline (I was confused for a minute thinking `\a` was a bell character...), but I think it would be much cleaner and more understandable to put the two lines in separate divs (or whatever block feels right).

(Also, attaching this newline to the last contributor means that works with >3 contributors have the "+ X" on the next line.)

app/components/moderation-list-row/component.js
```diff
@@ -39,35 +39,56 @@ export default Component.extend({
...
-    statusTimeDate: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
+    // date of preprint submission
+    submitDate: computed('submission.dateCreated', 'gtDay', function() {
+        return this.get('gtDay') ?
+            moment(this.get('submission.dateCreated')).format('MMMM DD, YYYY') :
+            moment(Math.min(Date.parse(this.get('submission.dateCreated')), Date.now())).fromNow();
+    }),
+
+    // translations for accepted and rejected records
+    acceptedRejectedStatusTimeDate: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
```
aaxelb: Similar to below, try to name things based on what they are or do. This is a description of the latest action made by a moderator, so maybe something like `moderatorActionLabel`?

app/components/moderation-list-row/component.js
aaxelb: Also, make sure the dependencies match what the property uses. It doesn't use `submission.reviewsState` or `noActions`, but does use `submitDate`.

app/components/moderation-list-row/component.js
```diff
@@ -39,35 +39,56 @@ export default Component.extend({
...
+        const labels = ACTION_LABELS[status][timeWording];
+        return i18n.t(labels, { timeDate: this.get('acceptedRejectedtDate'), moderatorName: this.get('latestActionCreator') });
+    }),
+
+    // translations for pending records
+    pendingStatusTimeDate: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
+        const i18n = this.get('i18n');
+        const dayValue = this.get('gtDay') ? 'gtDay' : 'ltDay';
+        const labels = ACTION_LABELS.pending[dayValue];
+        return i18n.t(labels, { timeDate: this.get('submitDate') });
```
aaxelb: It looks like this is the only place `submitDate` is used, is there a reason it's a computed property instead of a local variable? Computed properties are useful, but can make it harder to see how information flows.

app/components/moderation-list-row/component.js
```diff
@@ -39,35 +39,56 @@ export default Component.extend({
...
     }),
 
+    // check if it has been more than a day since preprint last transition
     gtDay: computed('submission.dateLastTransitioned', function() {
         return moment().diff(this.get('submission.dateLastTransitioned'), 'days') > 1;
     }),
 
-    relevantDate: computed('submission.dateLastTransitioned', 'gtDay', function() {
+    // date of preprint acceptance or rejection
+    acceptedRejectedtDate: computed('submission.dateLastTransitioned', 'gtDay', function() {
```
aaxelb: Typo: "Rejectedt"

app/components/moderation-list-row/component.js
```diff
@@ -39,35 +39,56 @@ export default Component.extend({
...
         return this.get('gtDay') ?
             moment(this.get('submission.dateLastTransitioned')).format('MMMM DD, YYYY') :
             moment(Math.min(Date.parse(this.get('submission.dateLastTransitioned')), Date.now())).fromNow();
     }),
 
-    // translations
-    statusTimeDate: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
+    // date of preprint submission
+    submitDate: computed('submission.dateCreated', 'gtDay', function() {
+        return this.get('gtDay') ?
```
aaxelb: Using `gtDay` here seems wrong; it looks at `submission.dateLastTransitioned`, not `submission.dateCreated`. So if something was submitted more than a day ago, but accepted less than a day ago, this will format the date wrong.

Now that we have two dates displayed, it might make sense to remove the `gtDay` property and just do that calculation in each place we need it (assuming that can be done without duplicated code).

app/locales/en/translations.js
```diff
@@ -212,14 +212,14 @@ export default {
...
-                submittedOn: 'submitted on {{timeDate}} by',
-                submitted: 'submitted {{timeDate}} by',
-                acceptedOn: 'accepted on {{timeDate}} by {{moderatorName}}',
-                accepted: 'accepted {{timeDate}} by {{moderatorName}}',
-                acceptedAutomaticallyOn: 'accepted automatically on {{timeDate}}',
-                acceptedAutomatically: 'accepted automatically {{timeDate}}',
-                rejectedOn: 'rejected on {{timeDate}} by {{moderatorName}}',
-                rejected: 'rejected {{timeDate}} by {{moderatorName}}',
+                submittedOn: 'Submitted on {{timeDate}} by {{contributorName}}',
+                submitted: 'Submitted {{timeDate}} by {{contributorName}}',
```
aaxelb: `{{contributorName}}` isn't ever used, it's just left blank.

tests/integration/components/moderation-list-row/component-test.js
```diff
@@ -25,30 +25,38 @@ test('it renders moderation-list-row accepted with actions', function(assert) {
...
+            contributors: [{ users: { fullName: 'Viper' } }, { users: { fullName: 'Oogway' } }],
+        }),
     });
     this.render(hbs`{{moderation-list-row submission=submission}}`);
     assert.ok(this.$('[data-status=accepted]').length);
     assert.notOk(this.$('[data-status=pending]').length);
     assert.notOk(this.$('[data-status=rejected]').length);
 
-    assert.equal(this.$('[data-status=accepted]').text().trim(), 'accepted on October 27, 2017 by Kung-fu Panda');
+    assert.equal(this.$('[data-status=accepted]').text().replace(/\s+/g, ' ').trim(), 'Submitted 2 months ago by' +
```
aaxelb: Does this test depend on the current date being about 2 months after October 27?

aaxelb (changes requested):
I had a few thoughts. Also, please update this to base on the newest `future-release/0.4.0`

app/components/moderation-list-row/component.js
```diff
@@ -55,19 +58,33 @@ export default Component.extend({
...
         const timeWording = this.get('noActions') ? `${dayValue}_automatic` : dayValue;
-        const labels = ACTION_LABELS[this.get('submission.reviewsState')][timeWording];
-        return i18n.t(labels, { timeDate: this.get('relevantDate'), moderatorName: this.get('latestActionCreator') });
+        const status = this.get('submission.reviewsState');
+        const labels = ACTION_LABELS[status][timeWording];
+        return i18n.t(labels, { timeDate: this.get('acceptedRejectedDate'), moderatorName: this.get('latestActionCreator') });
+    }),
+
+    // translations for submitted on label
+    submittedOnLabel: computed('gtDaySubmit', 'submission.dateCreated', function() {
```
aaxelb: It doesn't look like `gtDaySubmit` is anything this can depend on?

app/components/moderation-list-row/component.js
```diff
@@ -55,19 +58,33 @@ export default Component.extend({
...
+    acceptedRejectedDate: computed('submission.dateLastTransitioned', 'gtDay', function() {
         return this.get('gtDay') ?
             moment(this.get('submission.dateLastTransitioned')).format('MMMM DD, YYYY') :
             moment(Math.min(Date.parse(this.get('submission.dateLastTransitioned')), Date.now())).fromNow();
     }),
 
-    // translations
-    statusTimeDate: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
+    // translations for moderator action label
+    reviewedOnLabel: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
```
aaxelb: This should also depend on `acceptedRejectedDate` and `latestActionCreator`.

`gtDay` is now (even more) confusingly named. Do you think it would make sense to move it and `acceptedRejectedDate` into this property? It would closer parallel `submittedOnLabel` and reduce the number of interdependent properties.

app/components/moderation-list-row/template.hbs
```diff
@@ -7,17 +7,20 @@
...
                         {{/if}}
-                    {{else}}
-                        {{statusTimeDate}}
-                    {{/if}}
+                    </div>
+                    <div>
+                        {{#if (not-eq submission.reviewsState 'pending')}}
+                            {{reviewedOnLabel}}
+                        {{/if}}
+                    </div>
```
aaxelb: Nitpick: Move the `<div>` inside the `{{#if}}`

aaxelb (changes requested):
Just a few small things

app/components/moderation-list-row/component.js
sheriefvt: typo... was trying something and forgot to restore

app/components/moderation-list-row/component.js
```diff
@@ -55,19 +58,33 @@ export default Component.extend({
...
         const timeWording = this.get('noActions') ? `${dayValue}_automatic` : dayValue;
-        const labels = ACTION_LABELS[this.get('submission.reviewsState')][timeWording];
-        return i18n.t(labels, { timeDate: this.get('relevantDate'), moderatorName: this.get('latestActionCreator') });
+        const status = this.get('submission.reviewsState');
+        const labels = ACTION_LABELS[status][timeWording];
+        return i18n.t(labels, { timeDate: this.get('acceptedRejectedDate'), moderatorName: this.get('latestActionCreator') });
+    }),
+
+    // translations for submitted on label
+    submittedOnLabel: computed('gtDay', 'submission.dateCreated', function() {
```
aaxelb: This doesn't use `gtDay`.

app/components/moderation-list-row/component.js
```diff
@@ -55,19 +58,33 @@ export default Component.extend({
...
+    acceptedRejectedDate: computed('submission.dateLastTransitioned', 'gtDay', function() {
         return this.get('gtDay') ?
             moment(this.get('submission.dateLastTransitioned')).format('MMMM DD, YYYY') :
             moment(Math.min(Date.parse(this.get('submission.dateLastTransitioned')), Date.now())).fromNow();
     }),
 
-    // translations
-    statusTimeDate: computed('submission.reviewsState', 'gtDay', 'noActions', function() {
+    // translations for moderator action label
+    reviewedOnLabel: computed('submission.reviewsState', 'gtDay', 'noActions', 'acceptedRejectedDate', 'latestActionCreator', function() {
```
aaxelb: Thoughts on my comment from the last pass?
>`gtDay` is now (even more) confusingly named. Do you think it would make sense to move it and `acceptedRejectedDate` into this property? It would closer parallel `submittedOnLabel` and reduce the number of interdependent properties.

sheriefvt: can you clarify your comment, please? What do u mean by closer parallel?

aaxelb: I was suggesting getting rid of the `gtDay` and `acceptedRejectedDate` computed properties, and doing those computations inside `reviewedOnLabel` instead.

The main benefit would be reducing the amount of code and making it more readable.

Another benefit would be that `reviewedOnLabel` would have a more similar structure to `submittedOnLabel`. They would be closer to paralleling each other. That's good because they do almost the same thing, and doing similar things in a similar way makes it easier to understand and improve the code.

Do you think that would make sense to do?

sheriefvt: yes sure.. would look better

app/components/moderation-list-row/component.js
```diff
@@ -39,35 +39,45 @@ export default Component.extend({
...
-        const dayValue = this.get('gtDay') ? 'gtDay' : 'ltDay';
-        const timeWording = this.get('noActions') ? `${dayValue}_automatic` : dayValue;
-        const labels = ACTION_LABELS[this.get('submission.reviewsState')][timeWording];
-        return i18n.t(labels, { timeDate: this.get('relevantDate'), moderatorName: this.get('latestActionCreator') });
+        const gtDaySubmit = moment().diff(this.get('submission.dateCreated'), 'days') > 1;
+        const dayValue = gtDaySubmit ? 'gtDay' : 'ltDay';
+        const submitDate = gtDaySubmit ?
+            moment(this.get('submission.dateCreated')).format('MMMM DD, YYYY') :
+            moment(Math.min(Date.parse(this.get('submission.dateCreated')), Date.now())).fromNow();
+        const labels = ACTION_LABELS.pending[dayValue];
```
aaxelb: `pending` doesn't make sense as a name for this anymore, since this message is displayed for all states.

app/components/moderation-list-row/component.js
```diff
@@ -39,35 +39,45 @@ export default Component.extend({
...
 
-    gtDay: computed('submission.dateLastTransitioned', function() {
-        return moment().diff(this.get('submission.dateLastTransitioned'), 'days') > 1;
-    }),
-
-    relevantDate: computed('submission.dateLastTransitioned', 'gtDay', function() {
-        return this.get('gtDay') ?
+    // translations for moderator action label
+    reviewedOnLabel: computed('submission.{reviewsState,dateLastTransitioned}', 'noActions', 'latestActionCreator', function() {
+        const gtDay = moment().diff(this.get('submission.dateLastTransitioned'), 'days') > 1;
```
aaxelb: Ok, now that `reviewedOnLabel` and `submittedOnLabel` are more parallel, it's easier to see what's duplicated and we can DRY it up. Maybe a function like this would help?
```js
function friendlyDate(rawDate) {
    const date = moment(rawDate),
          now = moment();
    const gtDayAgo = now.diff(date, 'days') > 1;
    const formattedDate = gtDayAgo ?
        date.format('MMMM DD, YYYY') 
        moment.min(date, now).fromNow();
    return [formattedDate, gtDayAgo];
}
```

aaxelb (commented):
Looking good, now there's a clear opportunity to deduplicate some logic.

app/components/moderation-list-row/component.js
```diff
@@ -79,6 +83,14 @@ export default Component.extend({
         this.get('fetchData').perform();
     },
 
+    formattedDateLabel(rawDate) {
+        const gtDayAgo = moment().diff(rawDate, 'days') > 1;
+        const formattedDate = gtDayAgo ?
+            moment(rawDate).format('MMMM DD, YYYY') :
+            moment(Math.min(Date.parse(rawDate), Date.now())).fromNow();
```
aaxelb, line 90: If we're using `moment`, it would be nice to use it consistently instead of `Date`. I think it would make for more easily readable code.

app/components/moderation-list-row/template.hbs
```diff
@@ -7,16 +7,17 @@
             <div title={{submission.title}} class="submission-info">
                 <div class="submission-title">{{submission.title}}</div>
                 <div class="submission-body">
-                    {{#if (eq submission.reviewsState 'pending')}}
-                        {{statusTimeDate}}
+                    <div>
+                        {{submittedOnLabel}}
                         {{#each firstContributors as |contributor| ~}}
                             <li>{{contributor.users.fullName}}</li>
```
aaxelb, line 13: There was a [semi-relevant discussion](https://github.com/CenterForOpenScience/ember-osf-web/pull/60#discussion_r163326966) yesterday about shortened contributor lists on the OSF dashboard. It might be worth considering (at some point) updating reviews to match.

Essentially, say "X more" instead of "+ X", and use "and" when appropriate:
> by Sherif
by Sherif and Abram
by Sherif, Abram, and Lauren
by Sherif, Abram, Lauren, and 5 more

Not a blocker for this PR, though. @laurenbarker, thoughts?

aaxelb (approved):
Looks good! Couple comments, nothing blockingly important.

app/components/moderation-list-row/template.hbs
laurenbarker: That approach would be fine bit is a little out of scope for this PR. @sheriefvt if you'd like to make the suggested changes please put it in another PR (same ticket number is fine though)

PR #108, opened 2017-12-21
[Feature][MOD-181] Give warning on moderation detail page if reviewer input not submitted

sheriefvt:
Follow-up PR to rebase a previous PR #104 (will be closed) on future-release/0.4.0 

## Problem
If a reviewer accidentally exits the decision form we don't want them to navigate away from the page without making a decision.

## Goal
Give warning on moderation detail page if reviewer input (e.g. new/edit comment or change of decision) not submitted and reviewer tries to leave the page.

## Changes

- Update preprint-status-banner component
- Update preprint-detail page template, controller, and router
- Update translation file
- Add new addon ember-onbeforeunload for case 1 (see next)
- Add new unit tests

**Two cases**
- When user reloads or closes the page or navigate to OSF or another external app

- When user navigates to another route with-in the reviews app

## Ticket 
https://openscience.atlassian.net/browse/MOD-181


app/components/preprint-status-banner/component.js
```diff
@@ -227,11 +229,12 @@ export default Component.extend({
         return this.get('submission.reviewsState') !== this.get('decision');
     }),
 
+    notDefaultValues: computed('submission.reviewsState', 'decision', 'commentEdited', function() {
```
aaxelb: This doesn't use `submission.reviewsState`

app/components/preprint-status-banner/component.js
```diff
@@ -227,11 +229,12 @@ export default Component.extend({
...
+        return this.get('decision') !== ACCEPTED || this.get('commentEdited');
```
aaxelb: For rejected submissions, the "default" is `REJECTED`, so this doesn't prompt the user if they try to leave after changing from rejected to accepted. It might be good to save an `initialDecision` somewhere, which is the initial selected value.

app/components/preprint-status-banner/template.hbs
```diff
@@ -21,9 +21,9 @@
             {{/if}}
         </div>
         <div class="dropdown reviewer-feedback">
-            {{#bs-dropdown closeOnMenuClick=false as |dd|}}
+            {{#bs-dropdown onHide=(action 'closeReviewerFeedback') closeOnMenuClick=false as |dd|}}
                 {{#dd.button type="success" disabled=submission.reviewActions.isPending}}
-                    {{#if submission.reviewActions.isPending}}
+                    {{#if submission.actions.isPending}}
```
aaxelb: I think we want this `reviewActions` still

app/templates/preprints/provider/preprint-detail.hbs
```diff
@@ -20,7 +20,7 @@
     {{#if fetchData.isRunning}}
         <div class="preprint-status-component preprint-status-component--skeleton"></div>
     {{else}}
-        {{preprint-status-banner submission=preprint saving=savingAction submitDecision=(action 'submitDecision')}}
+        {{preprint-status-banner submission=preprint saving=savingAction userHasEnteredReview=userHasEnteredReview submitDecision=(action 'submitDecision')}}
```
aaxelb: It looks like `preprint-status-banner` doesn't actually use `userHasEnteredReview`, it just sets it to notify the controller of a change. Passing an action here would be a better/cleaner fit for that than a two-way binding. The component could use that action instead of `closeReviewerFeedback`.

aaxelb (changes requested):
Had a few small suggestions, and found a little bug.

app/components/preprint-status-banner/component.js
```diff
@@ -223,15 +226,16 @@ export default Component.extend({
         return this.get('reviewerComment').trim() !== this.get('initialReviewerComment');
     }),
 
-    decisionChanged: computed('submission.reviewsState', 'decision', function() {
-        return this.get('submission.reviewsState') !== this.get('decision');
+    decisionChanged: computed('decision', 'initialDecision', function() {
+        return this.get('initialDecision') !== this.get('decision');
+    }),
+
+    pendingStateNotChanged: computed('submission.reviewsState', 'decision', function() {
```
aaxelb: Is this property necessary? With `initialDecision` it seems like `decisionChanged`/`userActivity` are all you need.

app/controllers/preprints/provider/preprint-detail.js
```diff
@@ -124,6 +139,9 @@ export default Controller.extend({
 
             this._saveAction(action, filter);
         },
+        closeReviewerFeedback(userActivity) {
```
aaxelb: This name seems misleading -- it doesn't close anything.

app/components/preprint-status-banner/component.js
sheriefvt: There is a special case where submission status is pending, however, user decision is 'accept' on the feedback form. I used the pendingStateNotChanged property to check for it.

aaxelb: In that case, `initialDecision` and `decision` would both be `ACCEPTED`, so `decisionChanged` would be false, which is correct. Do we have to treat that case specially?

sheriefvt: Remember our discussion regarding the case where a user might change her/his decision (for pending preprints), then receives the warning, if the user restored decision to accept --> still should continue recieve the warning.

app/controllers/preprints/provider/preprint-detail.js
sheriefvt: what about 'setUserEnteredReview' ?

aaxelb: Sure, your call. As long as the name describes what the thing is or what the thing does.

app/components/preprint-status-banner/component.js
aaxelb: Ah, right. That doesn't need to be a special case, though -- you could set `userActivity = true` whenever the user clicks on a decision or enters text in the comment box. That might let you simplify some of the logic in this component, which is good because it looks like it's impossible to accept a pending preprint:

noaccept

app/components/preprint-status-banner/component.js
```diff
@@ -121,6 +123,10 @@ export default Component.extend({
 
     commentExceedsLimit: computed.gt('reviewerComment.length', COMMENT_LIMIT),
 
+    userActivity: computed.or('commentEdited', 'decisionUpdated'),
+
+    decisionUpdated: computed.and('decisionChanged', 'decisionValueToggled'),
```
aaxelb: It's confusing to have both `decisionUpdated` and `decisionChanged`, try to use more distinct names.

app/components/preprint-status-banner/component.js
```diff
@@ -257,20 +260,29 @@ export default Component.extend({
             this.set('decision', this.get('submission.reviewsState'));
             this.set('reviewerComment', this.get('initialReviewerComment'));
         },
+        closeFeedback() {
+            if (!this.get('pendingStateNotChanged') || this.get('userActivity')) {
```
aaxelb: `pendingStateNotChanged` doesn't exist anymore.

app/components/preprint-status-banner/component.js
```diff
@@ -121,6 +123,10 @@ export default Component.extend({
 
     commentExceedsLimit: computed.gt('reviewerComment.length', COMMENT_LIMIT),
 
+    userActivity: computed.or('commentEdited', 'decisionUpdated'),
```
aaxelb: Could this be simplified to `or('commentEdited', 'decisionValueToggled')`, getting rid of the intermediate `decisionUpdated`?

app/components/preprint-status-banner/component.js
```diff
@@ -257,20 +260,29 @@ export default Component.extend({
...
-
                 this.set('initialReviewerComment', comment);
                 this.set('reviewerComment', comment);
                 this.set('decision', this.get('submission.reviewsState'));
+                this.set('initialDecision', this.get('submission.reviewsState'));
             } else {
                 this.set('initialReviewerComment', '');
                 this.set('reviewerComment', '');
                 this.set('decision', ACCEPTED);
+                this.set('initialDecision', PENDING);
```
aaxelb: It looks like `initialDecision` is now just the same as `submission.reviewsState`, could we use that instead?

app/components/preprint-status-banner/template.hbs
```diff
@@ -21,7 +21,7 @@
             {{/if}}
         </div>
         <div class="dropdown reviewer-feedback">
-            {{#bs-dropdown closeOnMenuClick=false as |dd|}}
+            {{#bs-dropdown onHide=(action 'closeFeedback') closeOnMenuClick=false as |dd|}}
```
aaxelb: `closeFeedback` doesn't do much -- could this be `onHide=(action setUserEnteredReview userActivity)`?

app/templates/preprints/provider/preprint-detail.hbs
```diff
@@ -20,7 +20,7 @@
     {{#if fetchData.isRunning}}
         <div class="preprint-status-component preprint-status-component--skeleton"></div>
     {{else}}
-        {{preprint-status-banner submission=preprint saving=savingAction submitDecision=(action 'submitDecision')}}
+        {{preprint-status-banner submission=preprint saving=savingAction setUserEnteredReview=(action 'setUserEnteredReview') submitDecision=(action 'submitDecision')}}
```
aaxelb: You could use the [`mut` helper](https://emberjs.com/api/ember/2.16/classes/Ember.Templates.helpers) here, like `setUserEnteredReview=(action (mut userHasEnteredReview))`, which would let you delete `setUserEnteredReview`.

app/templates/preprints/provider/preprint-detail.hbs
```diff
@@ -161,4 +161,14 @@
         </div>
     </div>
 
+    {{#bs-modal open=showWarning onHide=(action 'hideWarning') as |modal|}}
+        {{#modal.header}}
+            <div class="warning-header"><i class="fa fa-exclamation-triangle"></i> {{t 'content.warning.header'}}</div>
+        {{/modal.header}}
+        {{#modal.body}}<div class="warning-body">{{t 'content.warning.body'}}</div>{{/modal.body}}
+        {{#modal.footer}}
+            {{#bs-button onClick=(action 'hideWarning') type="info"}}{{t 'content.warning.footer.stay'}}{{/bs-button}}
```
aaxelb: Minor: `onClick=(action modal.close)` would be better, so there's only one path out of the modal.

app/templates/preprints/provider/preprint-detail.hbs
```diff
@@ -161,4 +161,14 @@
         </div>
     </div>
 
+    {{#bs-modal open=showWarning onHide=(action 'hideWarning') as |modal|}}
```
aaxelb: This could be `onHide=(action (mut showWarning false))`, which would let you delete `hideWarning`.

aaxelb (commented):
The functionality looks good! I just had some small suggestions for reducing the amount of code.

app/controllers/preprints/provider/preprint-detail.js
```diff
@@ -110,6 +114,17 @@ export default Controller.extend({
...
         },
+        leavePage() {
+            const previousTransition = this.get('previousTransition');
+            if (previousTransition) {
+                this.set('userHasEnteredReview', false);
+                this.set('showWarning', false);
+                previousTransition.retry();
+            }
+        },
+        hideWarning() {
```
aaxelb: It doesn't look like `hideWarning` is used anymore.

aaxelb (commented):
Just one more small thing and this looks good to go 👍

app/routes/preprints/provider/preprint-detail.js
```diff
@@ -1,8 +1,12 @@
 import { isArray } from '@ember/array';
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+import ConfirmationMixin from 'ember-onbeforeunload/mixins/confirmation';
 
 
-export default Route.extend({
+export default Route.extend(ConfirmationMixin, {
+    theme: service(),
+    currentUser: service(),
```
aaxelb: Looks like `currentUser` isn't used anywhere.

aaxelb (dismissed):
Code is nice and clean, and works perfectly!

Just found one little unused injection, otherwise 💯

laurenbarker (approved):
Thanks for removing the unused injection! I'll merge once tests pass.

PR #110, opened 2018-01-24
[IN-59][Reviews] No redirection if going to a valid route with no permissions

adlius:
## Purpose

Currently, when a user who has no permission to the moderation page of a certain provider, we redirect them to `/reviews/forbidden`. This PR change the behavior so that we do not redirect, but still show a "Forbidden" page.

## Summary of Changes

In `routes/preprints/provider.js`, we no longer check for user permission but delegate that duty to the controller.
In `moderation.js` controller, a computed value `hasPermission` is added to check whether a user has permission for the current view.
In `templates/preprints/provider.hbs`, conditional is added to render "Forbidden" page if user has no permissions.

Check that if a user has no permission to a certain provider's moderation page (`reviews/preprints/<provider_id>`), instead of redirecting to `reviews/forbidden` to show the forbidden page, it should still show the "Forbidden" page but staying at `reviews/preprints/<provider_id>`.

## Ticket

https://openscience.atlassian.net/browse/IN-59

### DIFF
--- a/app/controllers/preprints/provider/moderation.js
+++ b/app/controllers/preprints/provider/moderation.js
@@ -26,10 +26,6 @@ export default Controller.extend(Analytics, moderationQueryParams.Mixin, {
     store: service(),
     theme: service(),
 
-    hasPermission: computed('model', function () {
-        return this.get('model.permissions').includes('view_submissions');
-    }),
-
     actions: {
         statusChanged(status) {
             this.resetQueryParams(['page']);
--- a/app/controllers/preprints/provider/moderation.js
+++ b/app/controllers/preprints/provider/moderation.js
@@ -1,3 +1,4 @@
+import { computed } from '@ember/object';
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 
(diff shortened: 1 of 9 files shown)


app/controllers/preprints/provider/moderation.js
```diff
@@ -6,7 +7,6 @@ import { task } from 'ember-concurrency';
 
 import Analytics from 'ember-osf/mixins/analytics';
 
-
```
laurenbarker, line 9: Not a big deal but the two spaces after the imports is intentional to make it easier to scan.

app/controllers/preprints/provider/moderation.js
```diff
@@ -26,6 +26,10 @@ export default Controller.extend(Analytics, moderationQueryParams.Mixin, {
     store: service(),
     theme: service(),
 
+    hasPermission: computed('model', function () {
```
laurenbarker: If this is defined in `moderation.js`, what happens if the user goes directly to settings or the detail page without permissions?

laurenbarker (changes requested):
Tiny thing and a question 🍔

PR #111, opened 2018-01-29
[IN-57][Fix] Use primary color for button

laurenbarker:
## Purpose
To be consistent with default modals on browsers, move the primary button to the right.

## Changes
Change the color of the right button to primary and use default color for the left.

## Side Effects
None


laurenbarker:
Tests passed on local travis: https://travis-ci.org/laurenbarker/ember-osf-reviews

PR #115, opened 2018-02-19
[IN-179][Reviews] Fix URL for download button

baylee-d:
## Purpose

Whenever a user tries to download a preprint, they are taken to an error page.  The user shouldn't be taken to an error page when doing this (and shouldn't be redirected at all).

## Changes

- Change the download URL to include `/preprints/{provider}`
  (before was `{window.origin}/guid/download`.  Now is `{window.origin}/preprints/{provider}/guid/download`

## QA Notes

This will affect anywhere the user can download the file on the `file-detail` page.  Check that the user can download before/after/approved/rejected moderation and that they don't get any errors.

## Ticket

https://openscience.atlassian.net/browse/IN-179

### DIFF
--- a/app/controllers/preprints/provider/preprint-detail.js
+++ b/app/controllers/preprints/provider/preprint-detail.js
@@ -27,6 +27,7 @@ export default Controller.extend({
     i18n: service(),
     toast: service(),
     store: service(),
+    theme: service(),
 
     queryParams: { chosenFile: 'file' },
 
@@ -64,6 +65,7 @@ export default Controller.extend({
         const { location: { origin } } = window;
         return [
             origin,
+            this.get('theme.id') !== 'osf' ? `preprints/${this.get('theme.id')}` : null,
             this.get('preprint.id'),
             'download',
         ].filter(part => !!part).join('/');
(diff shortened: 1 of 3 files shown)


laurenbarker (dismissed):
Checked out and tested but the functionality still seems to be broken. I submitted to a pre-moderation preprint service and the moderator couldn't download the preprint (it's possible it was originally tested with a superuser or something).

Also there are some failing tests 🐞

baylee-d:
Fixed the functionality to work for admins and moderators via this PR: https://github.com/CenterForOpenScience/osf.io/pull/8188

app/controllers/preprints/provider/preprint-detail.js
```diff
@@ -64,6 +65,7 @@ export default Controller.extend({
         const { location: { origin } } = window;
         return [
             origin,
+            this.get('theme.id') !== 'osf' ? `preprints/${this.get('theme.id')}` : null,
```
mfraezz, line 68: For future reference, this route is defined [here](https://github.com/CenterForOpenScience/osf.io/blob/76fb0c739f4cdabf03b5bfd2bc63d83b1c2d4796/website/routes.py#L279-L290) in `osf.io` code

mfraezz (approved):
LGTM

mfraezz:
Hotfixed in https://github.com/CenterForOpenScience/ember-osf-reviews/commit/8db582fa6e90475c4f47fc9a5285c3b82afffd08, consider this merged

PR #116, opened 2018-02-23
[IN-233][Reviews] Update to use latest osf-style with navbar changes

baylee-d:
## Purpose

New navbar changes have been merged into `osf-style`.  We should update to use the newest version of the navbar.

## Summary of Changes

- Update `osf-style` dependency to point at the latest version

Should use the newest navbar with flexbox

## Ticket

https://openscience.atlassian.net/browse/IN-95


alexschiller (approved):
James is looking into getting osf-style released, but that could be a bit since there are other things in dev. Will fix later. Merging.

PR #122, opened 2018-03-09
[IN-206][Reviews] Fix warning on preprint download 

fabmiz:
## Purpose
 
Fix issue where the user gets a warning modal when attempting
to download preprint after a decision has been submitted.

This issue was caused by `userHasEnteredReview` being left true on decision submission.

## Summary of Changes

- Set `userHasEnteredReview` to `false` on `submitDecision` action, hence preventing
the `showWarning` to be set to `true` in `willTransition` action.
- Updated tests

On the preprint detail page:

- Change the reviewer's decision
- Without submitting the form, attempt to click on any link on the page; make sure you are prevented from navigating forward and shown the warning modal. On the modal, hit `stay` or `leave` and see if they behave as expected.
- Back to the preprint detail page, modify decision, and submit. You should not get the warning modal.
- After submitting a decision, attempt to download the preprint.

## Ticket

[IN-206](https://openscience.atlassian.net/browse/IN-206)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] testable and includes test(s)
- [ ] changes described in `CHANGELOG.md`

### DIFF
--- a/app/controllers/preprints/provider/preprint-detail.js
+++ b/app/controllers/preprints/provider/preprint-detail.js
@@ -126,7 +126,7 @@ export default Controller.extend({
         },
         submitDecision(trigger, comment, filter) {
             this.toggleProperty('savingAction');
-
+            this.set('userHasEnteredReview', false);
             const action = this.store.createRecord('review-action', {
                 actionTrigger: trigger,
                 target: this.get('preprint'),
--- a/app/controllers/preprints/provider/preprint-detail.js
+++ b/app/controllers/preprints/provider/preprint-detail.js
@@ -126,7 +126,6 @@ export default Controller.extend({
         },
         submitDecision(trigger, comment, filter) {
             this.toggleProperty('savingAction');
-            this.set('userHasEnteredReview', false);
             const action = this.store.createRecord('review-action', {
                 actionTrigger: trigger,
                 target: this.get('preprint'),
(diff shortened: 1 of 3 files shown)


app/controllers/preprints/provider/preprint-detail.js
```diff
@@ -126,7 +126,7 @@ export default Controller.extend({
         },
         submitDecision(trigger, comment, filter) {
             this.toggleProperty('savingAction');
-
+            this.set('userHasEnteredReview', false);
```
baylee-d: Would it be better to change this property after the review has been successfully saved?

baylee-d (commented):
Just a quick question, but other than that it looked gtm

app/controllers/preprints/provider/preprint-detail.js
fabmiz: `savingAction` is the property keeping track of that, right? I do not see much benefits from that change. (in case of promise failure) carrying over state changes after the corresponding action have completed might cause issues if the property is used in some other ways.

baylee-d: My only concern would be if the user tries to save their review and failed.  As I'm understanding it, `userHasEnteredReview` would still be changed to `false` even though the user's review wasn't saved, and the state of dirty fields for transitions could get a bit messed up because it's using that property as it's base.

fabmiz: @baylee-d, I see what you mean. Thanks.

alexschiller (approved):
LGTM

PR #123, opened 2018-03-14
[IN-6][Reviews] Add moderator management page

laurenbarker:
## TODO
- [x] get adding a new moderator working
- [x] get removing moderator working
- [x] get edit controls working
- [x] get adding a moderator by email working
- [x] fix row loading screen
- [x] move all text to translation file
- [x] update the error messages
- [x] add pagination controls
- [x] sort results alphabetically (backend)
- [x] fix linting errors
- [x] disable users in search results who are already moderators
- [x] disable everything for non-admin moderators
- [x] force refresh if a user is added or if pagination changes
- [x] add tests

## Purpose
Allow preprint providers to add and remove moderators using the moderation app.

## Depends on
* https://github.com/CenterForOpenScience/osf.io/pull/8186 (released)
* https://github.com/CenterForOpenScience/ember-osf/pull/370

## Changes
* Added "Moderators" tab
* Added "Moderators" link on dashboard
* Changed "Moderation" tab to "Submissions"
* Changed "Moderation" to "Submissions" on dashboard
* Made the tabs a little better shaped
* Added all of the moderator management functionality

moderator-management

## Testing
Should look and act like [prototype](https://marvelapp.com/7hg387e/) with a few exceptions:
* When a new user is selected it keeps the input field
* Using header row instead of page title
* "Add new moderator" button is now just "Add"
* Moderators (as opposed to admin) cannot see the Permissions/Add/Remove functionality
* "Remove" instead of "Delete" moderators

Known oddities:
* You **can't add super users as moderators through the interface** (they have access though)
* We don't have a way of knowing the order that the users were added so alphabetical will have to do. This means that users added at the end might be in a different place if the page is reloaded.

## Ticket

### DIFF
--- a/app/components/moderator-list-add/component.js
+++ b/app/components/moderator-list-add/component.js
@@ -44,10 +44,14 @@ export default Component.extend(Validations, {
     unregisteredUserName: '',
     unregisteredUserEmail: '',
     selectedUnregisteredUser: false,
-    roleLabel: 'Role',
     role: '',
 
     isFormValid: computed.alias('validations.isValid'),
+
+    roleLabel: computed('role', function() {
+        const roleOption = this.get('roleOptions').findBy('role', this.get('role'));
+        return roleOption ? roleOption.label : 'Role';
+    }),
 
     selectedUserId: computed('selectedUser', 'unregisteredUserEmail', function() {
         return this.get('selectedUser.id') || this.get('unregisteredUserEmail');
@@ -64,11 +68,6 @@ export default Component.extend(Validations, {
     actions: {
         roleChanged(role) {
             this.set('role', role);
-            for (const value of this.get('roleOptions')) {
-                if (value.role === role) {
-                    this.set('roleLabel', value.label);
-                }
-            }
         },
         cancel() {
             this.setProperties({
@@ -97,29 +96,19 @@ export default Component.extend(Validations, {
         yield timeout(DEBOUNCE_MS);
 
         try {
-            // Hack to disable search results for users who are already moderators
-            // ember-data doesn't like when you manipulate the output
-            // ember-power-select expects disabled=true on disabled results
-            const users = yield $.ajax({
-                type: 'GET',
-                url: `${config.OSF.apiUrl}/${config.OSF.apiNamespace}/users`,
-                headers: {
-                    ACCEPT: 'application/vnd.api+json; version=2.6',
+            const users = yield this.get('store').query('user', {
+                filter: {
+                    'full_name,given_name,middle_names,family_name': query,
                 },
-                data: {
-                    filter: {
-                        'full_name,given_name,middle_names,family_name': query,
-                    },
-                    page: {
-                        size: 15,
-                    },
+                page: {
+                    size: 15,
                 },
             });
 
-            const usersFullNames = users.data.map((user) => {
+            const usersFullNames = users.content.map((user) => {
                 const userData = {
-                    fullName: user.attributes.full_name,
-                    profileImage: user.links.profile_image,
+                    fullName: user.__data.fullName,
+                    profileImage: user.__data.links.profile_image,
                     id: user.id,
                 };
                 if (this.get('moderatorIds').indexOf(user.id) > -1) {
--- a/app/components/moderator-list-add/component.js
+++ b/app/components/moderator-list-add/component.js
@@ -1,0 +1,57 @@
+import { computed } from '@ember/object';
+import { inject as service } from '@ember/service';
+import Component from '@ember/component';
+
+import { task, timeout } from 'ember-concurrency';
+
+
+const DEBOUNCE_MS = 250;
+
+export default Component.extend({
+    i18n: service(),
+    store: service(),
+
+    selectedUser: '',
+
+    role: computed('moderator.permissionGroup', 'roleOptions', function() {
+        for (const value of this.get('roleOptions')) {
+            if (value.role === this.get('moderator.permissionGroup')) {
+                return value.label;
+            }
+        }
+        return 'Role';
+    }),
+
+    disableSave: computed('role', 'user', function() {
+        return this.get('role') === 'Role' || !this.get('selectedUser');
+    }),
+
+    actions: {
+        roleChanged(role) {
+            this.set('role', role);
+        },
+        cancel() {
+            this.setProperties({
+                editingModerator: false,
+                addingNewModerator: false,
+                role: 'Role',
+                selectedUser: '',
+            });
+        },
+    },
+
+    searchUsers: task(function* (query) {
+        yield timeout(DEBOUNCE_MS);
+        try {
+            const users = yield this.get('store').query('user', {
+                filter: {
+                    'full_name,given_name,middle_names,family_name': query,
+                },
+                size: 20,
+            });
+            return users;
+        } catch (e) {
+            this.get('toast').error(this.get('i18n').t('submit.search_contributors_error'));
+        }
+    }).restartable(),
+});
--- a/app/components/moderator-list-add/component.js
+++ b/app/components/moderator-list-add/component.js
@@ -5,8 +5,6 @@ import Component from '@ember/component';
 import { task, timeout } from 'ember-concurrency';
 import { validator, buildValidations } from 'ember-cp-validations';
 
-import config from '../../config/environment';
-
 
 const DEBOUNCE_MS = 250;
 
--- a/app/components/moderator-list-add/component.js
+++ b/app/components/moderator-list-add/component.js
@@ -8,8 +8,9 @@ import { validator, buildValidations } from 'ember-cp-validations';
 
 const DEBOUNCE_MS = 250;
 
-/* Validations for adding unregistered contributor form.  fullName must be present
-and have three letters, and the username (email) must be present and of appropriate format.
+/* Validations for adding unregistered contributor form.
+* fullName must be present and have at least three characters but not more than 255
+* username (email) must be present and of appropriate format.
 */
 const Validations = buildValidations({
     fullName: {
@@ -18,6 +19,7 @@ const Validations = buildValidations({
             validator('presence', true),
             validator('length', {
                 min: 3,
+                max: 255,
             }),
         ],
     },
@@ -109,7 +111,7 @@ export default Component.extend(Validations, {
                     profileImage: user.__data.links.profile_image,
                     id: user.id,
                 };
-                if (this.get('moderatorIds').indexOf(user.id) > -1) {
+                if (this.get('moderatorIds').includes(user.id)) {
                     userData.disabled = true;
                 }
                 return userData;
--- a/app/components/moderator-list-add/component.js
+++ b/app/components/moderator-list-add/component.js
@@ -5,6 +5,8 @@ import Component from '@ember/component';
 import { task, timeout } from 'ember-concurrency';
 import { validator, buildValidations } from 'ember-cp-validations';
 
+import config from '../../config/environment';
+
 
 const DEBOUNCE_MS = 250;
 
@@ -36,6 +38,7 @@ const Validations = buildValidations({
 export default Component.extend(Validations, {
     i18n: service(),
     store: service(),
+    theme: service(),
 
     selectedUser: '',
     unregisteredUserName: '',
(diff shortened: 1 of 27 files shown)


app/components/moderator-list-add/component.js
```diff
@@ -0,0 +1,136 @@
...
+            }),
+        ],
+    },
+    email: {
+        description: 'Email',
+        validators: [
+            validator('presence', true),
+            validator('format', {
+                type: 'email',
+                regex: /^[-a-z0-9~!$%^&*_=+}{'?]+(\.[-a-z0-9~!$%^&*_=+}{'?]+)*@([a-z0-9_][-a-z0-9_]*(\.[-a-z0-9_]+)*\.(aero|arpa|biz|com|coop|edu|gov|info|int|mil|museum|name|net|org|pro|travel|mobi|[a-z][a-z])|([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}))(:[0-9]{1,5})?$/i,
```
aaxelb, line 32: So IP hosts are fine, but you'll reject `anyone@dinosaurs.sexy`?

app/components/moderator-list-add/component.js
```diff
@@ -0,0 +1,136 @@
...
+            }
+        },
+    },
+
+    searchUsers: task(function* (query) {
+        yield timeout(DEBOUNCE_MS);
+
+        try {
+            // Hack to disable search results for users who are already moderators
+            // ember-data doesn't like when you manipulate the output
```
aaxelb: Couldn't you take the ember-data output and do the `map` you have below?
```js
const users = yield store.query('user', ...);
return users.map(user => {
  const userData = { ... };
  if (...) {
    userData.disabled = true;
  }
  return userData;
});
```
 Or is there something I'm missing?

app/components/moderator-list-add/template.hbs
```diff
@@ -0,0 +1,96 @@
+<div class="moderators-row-container p-h-md p-v-sm">
+    <div class="moderator-name">
+        {{#if fetchData.isRunning}}
```
aaxelb: `searchUsers`?

tests/integration/components/moderator-list-row/component-test.js
```diff
@@ -0,0 +1,93 @@
...
+moduleForComponent('moderator-list-row', 'Integration | Component | moderator-list-row', {
+    integration: true,
+
+    beforeEach() {
+        this.register('service:store', storeStub);
+        this.inject.service('store');
+    },
+});
+
+test('it renders moderator-list-row admin view', function(assert) {
```
aaxelb, line 28: Tests!?

app/components/moderator-list-add/component.js
```diff
@@ -0,0 +1,136 @@
...
+    }),
+
+    showInviteForm: computed('selectedUser', 'selectedUnregisteredUser', function() {
+        return !this.get('selectedUser') && !this.get('selectedUnregisteredUser');
+    }),
+
+    actions: {
+        roleChanged(role) {
+            this.set('role', role);
+            for (const value of this.get('roleOptions')) {
```
aaxelb: Equivalent, using `findBy`:
```js
this.set('roleLabel', this.get('roleOptions').findBy('role', role).label);
```
(but also maybe `roleLabel` should be a CP)

app/components/moderator-list-row/component.js
```diff
@@ -0,0 +1,81 @@
...
+    i18n: service(),
+    store: service(),
+
+    removeConfirmation: false,
+    editConfirmation: false,
+    roleLabel: 'Role',
+    role: computed('moderator.permissionGroup', function() {
+        for (const value of this.get('roleOptions')) {
+            if (value.role === this.get('moderator.permissionGroup')) {
+                this.set('roleLabel', value.label);
```
aaxelb: Computed properties with side effects are generally a bad idea, since ember will do all it can to avoid running it. Seems like it would make more sense for `roleLabel` to be a CP itself?

app/components/moderator-list-row/component.js
```diff
@@ -0,0 +1,81 @@
...
+            this.setProperties({
+                editingModerator: false,
+                removeConfirmation: false,
+                editConfirmation: false,
+            });
+        },
+    },
+
+    _setRole(role) {
+        this.set('role', role);
```
aaxelb: `role` is a computed property, it's confusing to also set it...

app/components/moderator-list-row/component.js
```diff
@@ -0,0 +1,81 @@
...
+        for (const value of this.get('roleOptions')) {
```
aaxelb: Could be shorter with `findBy`, again.

app/controllers/preprints/provider/moderators.js
```diff
@@ -0,0 +1,216 @@
...
+            const response = yield this.get('store').query('moderator', {
+                provider: this.get('theme.provider.id'),
+                filter: { id },
+            });
+            const moderatorInstance = response.get('firstObject');
+            moderatorInstance.provider = this.get('theme.provider.id');
+
+            yield moderatorInstance.destroyRecord({ adapterOptions: { provider: this.get('theme.provider.id') } });
+
+            const allModerators = yield this.get('store').peekAll('moderator');
```
aaxelb: `peekAll` is synchronous; the `yield` does nothing.

app/controllers/preprints/provider/moderators.js
```diff
@@ -0,0 +1,216 @@
...
+            }
+        },
+    },
+
+    setup({ queryParams }) {
+        this.set('moderatorIds', []);
+        this.set('roleOptions', [
+            {
+                role: 'admin',
+                label: 'Admin',
```
aaxelb: i18n?

app/styles/modules/_provider.scss
```diff
@@ -1,3 +1,4 @@
 @import 'provider/settings';
 @import 'provider/setup';
+@import 'provider/moderators';
```
aaxelb, line 3: Highly recommend using ember-component-css [for route stylesheets too](https://github.com/ebryn/ember-component-css#usage-with-routes), rather than this `modules` setup.

Also, pod routes.

app/controllers/preprints/provider/moderators.js
```diff
@@ -0,0 +1,216 @@
...
+            provider: this.get('theme.provider.id'),
+        });
+        const moderatorIds = moderators.map(moderator => moderator.id);
+        this.set('moderatorIds', moderatorIds);
+    }),
+
+    deleteModerator: task(function* (id) {
+        try {
+            // queryRecord doesn't play nice with the API
+            // findRecord with delete won't accept a payload
```
aaxelb: Could you pass `moderator` instead of `moderator.id` and do `moderator.destroyRecord()` right off the bat? Or is it important to double-check that it wasn't already deleted?

aaxelb (commented):
🍰

app/components/moderator-list-add/component.js
laurenbarker: Oh lol I didn't write this I got it from https://github.com/CenterForOpenScience/ember-osf-preprints/blob/develop/app/components/unregistered-contributor-form.js#L27

app/components/moderator-list-add/component.js
laurenbarker: So I thought the same thing but Ember did not agree. I don't think I completely removed the ember stuff when I was mapping though so I'll try that.

app/controllers/preprints/provider/moderators.js
laurenbarker: Oops, good catch!

app/controllers/preprints/provider/moderators.js
laurenbarker: I'm not sure if I tried that -- fingers crossed 🤞

app/components/moderator-list-row/component.js
laurenbarker: Yeah...the issue is I need role to be computed on the initial load and then set after that. I tried having it in `didRecieveAttrs` but that was triggered every time role changed. I think it would work in `init` though. Agree that it should not be a computed.

app/components/moderator-list-add/template.hbs
laurenbarker: Hm this might be bad copy pasta...

app/components/moderator-list-row/component.js
aaxelb: So could it be a [oneWay](https://www.emberjs.com/api/ember/3.0/functions/@ember%2Fobject%2Fcomputed/oneWay) (or its more understandable alias, `reads`)?
```js
role: reads('moderator.permissionGroup')
```

app/styles/modules/_provider.scss
laurenbarker: Little out of scope but will create a [ticket](https://openscience.atlassian.net/browse/IN-219) because I agree with you.

app/controllers/preprints/provider/moderators.js
```diff
@@ -0,0 +1,200 @@
...
+                this.get('results.moderators').removeObject(moderatorInstance);
+            }
+
+            yield this.get('fetchAdmin').perform();
+            return true;
+        } catch (e) {
+            this.get('toast').error(this.get('i18n').t('moderators.deleteModeratorError'));
+            return false;
+        } finally {
+            this.set('editingModerator', false);
```
alexschiller: Edgy: 
screen shot 2018-04-20 at 11 37 54 am
Should the moderators be reloaded or `moderatorIds` be updated to prevent deleted mods from being disabled in search?

app/components/moderator-list-add/template.hbs
```diff
@@ -0,0 +1,90 @@
...
+        {{#if (not selectedUnregisteredUser)}}
+            {{#power-select
+                disabled=loadingModerators
+                options=users
+                search=(perform searchUsers)
+                placeholder='Search by name'
+                selected=selectedUser
+                onchange=(action (mut selectedUser)) as |user|}}
+                <img class="m-xs" src="{{user.profileImage}}" height=25 width=25 alt="{{user.fullName}} gravatar">
+                <span>{{user.fullName}}</span>
```
alexschiller: Should this truncate? my full name (albeit all caps) goes onto a second line.
screen shot 2018-04-20 at 12 06 58 pm

app/components/moderator-list-row/template.hbs
```diff
@@ -0,0 +1,75 @@
...
+            <img class="visible-xs" src="{{user.links.profile_image}}" height=25 width=25 alt="{{moderator.fullName}} gravatar">
+            <div class="name p-l-sm">
+                <a href="{{user.links.html}}">{{moderator.fullName}}</a>
+            </div>
+        {{/if}}
+    </div>
+    {{#if isAdmin}}
+        {{#bs-dropdown classNames="pull-right role-dropdown" as |dd|}}
+            {{#dd.button classNames="btn btn-default dropdown-button" disabled=disableRoleSelect}}
+                {{roleLabel}}
```
alexschiller: So it looks like mods only see the mod name list-- should they at least be able to see what role each person has (disabled button maybe)?  Seems they can see that data via the api, so it should be possible. If desirable, this can be a deferred improvement.

User story: I'm a moderator for a large preprint service and I want to be removed as a mod/upgraded to an admin but don't know who to contact.

app/components/moderator-list-add/template.hbs
```diff
@@ -0,0 +1,90 @@
...
+                    <button class="btn" {{action 'roleChanged' roleOption.role}}>
+                        {{roleOption.label}}
+                        <i class='{{if (eq role roleOption.role) 'fa fa-check selected' 'not-selected'}}'></i>
+                    </button>
+                {{/menu.item}}
+            {{/each}}
+        {{/dd.menu}}
+    {{/bs-dropdown}}
+    <div class="row-controls">
+        <button class="btn btn-default" {{action 'cancel'}} disabled={{addModerator.isRunning}}>
```
alexschiller, line 80: Cancels get gnarly on small screen sizes... one of those, "probably not a priority, but..."
screen shot 2018-04-20 at 12 21 18 pm
screen shot 2018-04-20 at 11 38 31 am

app/components/moderator-list-add/component.js
```diff
@@ -0,0 +1,123 @@
...
+/* Validations for adding unregistered contributor form.  fullName must be present
+and have three letters, and the username (email) must be present and of appropriate format.
+*/
+const Validations = buildValidations({
+    fullName: {
+        description: 'Full name',
+        validators: [
+            validator('presence', true),
+            validator('length', {
+                min: 3,
```
alexschiller, line 21: This should probably have a max as well, db max is set to 255

app/locales/en/translations.js
```diff
@@ -230,6 +247,14 @@ export default {
                 rejected: 'Rejected {{timeDate}} by {{moderatorName}}',
             },
         },
+        moderatorListAdd: {
+            inviteText: 'invite by email',
```
alexschiller: stupid minor; when this text is highlighted it looks really cramped. Not sure if there is an easy fix. If you know one, do it, if not, leave it.
screen shot 2018-04-20 at 11 33 11 am

app/components/moderator-list-add/component.js
```diff
@@ -0,0 +1,123 @@
...
+                },
+            });
+
+            const usersFullNames = users.content.map((user) => {
+                const userData = {
+                    fullName: user.__data.fullName,
+                    profileImage: user.__data.links.profile_image,
+                    id: user.id,
+                };
+                if (this.get('moderatorIds').indexOf(user.id) > -1) {
```
alexschiller: I think this can use the es6 Array.includes
```javascript
if (this.get('moderatorIds').includes(user.id)) {
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

alexschiller (changes requested):
Couple things to look into.
At a minimum, add max length validation for full names. Open to thoughts/new tickets for other things as appropriate.

app/components/moderator-list-row/template.hbs
laurenbarker: It was decided in demo (and then not commented anywhere) that moderators would only see a list of names. It's definitely not a bad idea for moderators to see the the role though.

I'll follow up with product to make sure I'm remembering the decision correctly.

alexschiller: :+1: Sounds good. It's a minor improvement, so nbd... that page just felt empty.

app/locales/en/translations.js
laurenbarker: I can make it a little better at least:
screen shot 2018-04-20 at 4 16 03 pm

alexschiller: Perfect ⭐️

app/components/moderator-list-add/template.hbs
laurenbarker: screen shot 2018-04-23 at 10 01 24 am

app/components/moderator-list-add/template.hbs
laurenbarker: Can scroll within the dropdown if super long names:
screen shot 2018-04-23 at 10 03 22 am

alexschiller (changes requested):
As expected, conflicts ⚔️

alexschiller (approved):
🍰

PR #125, opened 2018-03-26
[IN-168][Reviews] Add Reviews Notifications page

adlius:
## Purpose

This PR adds Notifications page to Reviews app.

## Summary of Changes

1. Added new component `notification-list`.
2. Added new component `notification-list-row`.
3. Changed `reviews-breadcrumbs/component.js` to capitalize the breadcrumb.
4. Added tests for new components and routes.

None.

## Testing Notes

This PR, once merged, plays with the previously added `/subscriptions` endpoint (https://openscience.atlassian.net/browse/IN-167) so that users are able to change their subscription.
Several things to test:
1. Make sure that changing the notification frequency actually works.
2. Make sure there is placeholder during initial loading of the page. Placeholder is not needed when you change the notification frequency.
3. Make sure that alerts are shown when updating notification fails.
4. Make sure that it looks good.

## Ticket

https://openscience.atlassian.net/browse/IN-168


app/components/notification-list-row/template.hbs
```diff
@@ -0,0 +1,23 @@
+<div class="flex-container">
+    <div class="notification-title">New pending submissions</div>
+    {{#bs-dropdown as |dd|}}
+        {{#dd.button class="dropdown"}} {{t currentSettingLabel}} <span class="caret"></span>{{/dd.button}}
```
laurenbarker: Did you end up adding the padding between the caret and the label? Should be able to add `p-l-xs` as a class on the span.

Also, I think that we should use font-awesome carets (`fa fa-caret-down`) instead of bootstrap's default caret since we use font awesome icons everywhere else. We should avoid having multiple types of the same icon on the site.

app/components/notification-list-row/template.hbs
```diff
@@ -0,0 +1,23 @@
...
+        {{#dd.menu classNames="notification-menu" as |menu|}}
+            {{#menu.item}}
```
laurenbarker: This could be DRY'd up a little bit by adding a list in the component. Something like:
```
// component.js
const notificationOptions = [
    {
        action: 'updateFrequency',
        label: this.get('i18n').t('components.notificationListRow.instant'),
        value: 'instant',
    },
    {...}
    {...}
];
```
```
// template.hbs
{{#each notificationOptions as |notificationOption|}}
    {{#menu.item}}
        <button class="btn" onclick={{action notificationOption.action notificationOption.value}}>
            {{notificationOption.label}}
        </button>
    {{/menu.item}}
{{/each}}
```

app/components/notification-list-row/template.hbs
```diff
@@ -0,0 +1,23 @@
...
+                </button>
+            {{/menu.item}}
+            {{#menu.item}}
+                <button class="btn" onclick={{action 'updateFrequency' 'none'}}>
+                    {{t frequencyLabel.never}}
+                </button>
+            {{/menu.item}}
+        {{/dd.menu}}
+    {{/bs-dropdown}}
+</div>
```
laurenbarker: Need a newline here (it's useful to set your editor to add newlines on save).

app/components/notification-list/template.hbs
```diff
@@ -0,0 +1,19 @@
+<!-- template-lint triple-curlies=false -->
```
laurenbarker: What is this for?

app/components/notification-list/template.hbs
```diff
@@ -0,0 +1,19 @@
...
+            <div class="notification-list-body">
+                {{#each subscriptions as |subscription|}}
+                    <hr>
+                    {{notification-list-row subscription=subscription}}
+                    <hr>
+                {{/each}}
+            </div>
+        </div>
+    </div>
+</div>
```
laurenbarker: Need a newline here too

app/templates/preprints/provider/notifications.hbs
```diff
@@ -0,0 +1,3 @@
+{{notification-list
+    subscriptions=model
+}}
```
laurenbarker: Newline here too

app/components/reviews-breadcrumbs/component.js
```diff
@@ -22,7 +26,7 @@ export default Component.extend({
             if (r.context === breadcrumbs[breadcrumbs.length - 1].context) continue;
 
             // Prefer model name or title, fall back to the route name
-            set(r, 'name', r.context.breadcrumbTitle || (r.context.get && r.context.get('name')) || r.part);
+            set(r, 'name', capitalize(r.context.breadcrumbTitle || (r.context.get && r.context.get('name')) || r.part));
```
laurenbarker: You can probably [capitalize the first letter using css](https://github.com/CenterForOpenScience/ember-osf-reviews/blob/develop/app/components/preprint-status-banner/style.scss#L55) instead of in the javascript.

app/routes/preprints/provider/notifications.js
```diff
@@ -0,0 +1,16 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default Route.extend({
+    theme: service(),
+
+    model() {
```
laurenbarker: Instead of loading the model in the route could you move it to the controller and use [ember-concurrency](http://ember-concurrency.com/docs/introduction/)? That way the whole route isn't blocked from loading and we can use skeleton screens to reduce the time to [first meaningful paint](https://developers.google.com/web/tools/lighthouse/audits/first-meaningful-paint).

app/routes/preprints/provider/notifications.js
```diff
@@ -0,0 +1,16 @@
...
+        const providerId = this.get('theme.id');
+        const subscriptionId = `${providerId}_new_pending_submissions`;
+        return this.get('store').query('subscription', {
```
laurenbarker: Are you only expecting one result here? If so you can use something more specific like `queryRecord` or `findRecord`.

app/components/notification-list/template.hbs
```diff
@@ -0,0 +1,19 @@
...
+    <div class="row">
+        <div class="notification-list col-md-8">
+            <div>
+                <h4 class="notification-list-heading">{{t 'components.notificationList.heading'}}</h4>
+                <span class="helper-text"> {{t 'components.notificationList.helperText'}} <a href={{serviceLinks.settings}}>{{t 'components.notificationList.userSettings'}}</a>.</span>
+            </div>
+
+            <div class="notification-list-body">
+                {{#each subscriptions as |subscription|}}
+                    <hr>
```
laurenbarker: `<hr>` tags shouldn't be used here -- see [Mozilla Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr):
> While it may still be displayed as a horizontal rule in visual browsers, this element is now defined in semantic terms, rather than presentational terms, so if you wish to draw a horizontal line, you should do so using appropriate CSS.

app/components/notification-list/template.hbs
```diff
@@ -0,0 +1,19 @@
+<!-- template-lint triple-curlies=false -->
+<div class="container">
+    <div class="row">
+        <div class="notification-list col-md-8">
+            <div>
+                <h4 class="notification-list-heading">{{t 'components.notificationList.heading'}}</h4>
+                <span class="helper-text"> {{t 'components.notificationList.helperText'}} <a href={{serviceLinks.settings}}>{{t 'components.notificationList.userSettings'}}</a>.</span>
```
laurenbarker: This would be more readable if it was on separate lines:
```
<span class="helper-text"> 
    {{t 'components.notificationList.helperText'}} 
    <a href={{serviceLinks.settings}}>
        {{t 'components.notificationList.userSettings'}}
    </a>.
</span>
```

app/components/notification-list-row/component.js
```diff
@@ -0,0 +1,34 @@
...
+        if (frequency === 'daily') {
+            return frequencyLabel.daily;
+        }
+        if (frequency === 'none') {
+            return frequencyLabel.never;
+        }
+    }),
+
+    actions: {
+        updateFrequency: (frequency) => {
```
laurenbarker: You should use [ember-concurrency](http://ember-concurrency.com/docs/introduction/) here since `subscription.save()` makes a request. Something like:

```
updateFrequency(task(function* (frequency) {
    try {
        const subscription = this.get('subscription');
        subscription.set('frequency', frequency);
        yield subscription.save();
    } catch (e) {
        this.get('toast').error(this.get('i18n').t('...'));
    }
});
```
Then in the template you could add a saving indicator by checking `updateFrequency.isRunning`.

app/components/notification-list-row/component.js
```diff
@@ -0,0 +1,34 @@
...
+    daily: 'components.notificationListRow.daily',
+    never: 'components.notificationListRow.never',
+};
+
+export default Component.extend({
+    i18n: service(),
+
+    currentSettingLabel: computed('subscription.frequency', function () {
+        const frequency = this.get('subscription.frequency');
+        if (frequency === 'instant') {
```
laurenbarker: This can also be DRY'd up if you use an array and [findBy](https://emberjs.com/api/ember/2.18/classes/EmberArray/methods/findBy?anchor=findBy) (h/t Abram for findBy).

laurenbarker (changes requested):
Nothing major just some minor refactors.

A few other things not captured in the other comments:
* Don't need `_notifications.scss` if there isn't anything in it. We shouldn't have any empty files lying around.
* Could you add a link to the notifications tab from the [dashboard](https://github.com/CenterForOpenScience/ember-osf-reviews/blob/develop/app/components/provider-links/template.hbs#L25)?

app/components/notification-list-row/component.js
```diff
@@ -1,34 +1,43 @@
...
+        try {
             subscription.set('frequency', frequency);
-            subscription.save();
-        },
-    },
+            if (subscription.changedAttributes()) {
+                yield subscription.save();
+            }
+        } catch (e) {
+            subscription.rollbackAttributes();
```
laurenbarker, line 39: Nice, I always forget this a thing

app/components/notification-list-row/component.js
```diff
@@ -1,34 +1,43 @@
...
-            const subscription = this.get('subscription');
+    updateFrequency: task(function* (frequency) {
+        const subscription = this.get('subscription');
+        try {
             subscription.set('frequency', frequency);
-            subscription.save();
-        },
-    },
+            if (subscription.changedAttributes()) {
+                yield subscription.save();
```
laurenbarker: What happens if this yield isn't hit? Does the task keep running?

app/components/notification-list/component.js
```diff
@@ -1,7 +1,24 @@
...
+import { inject as service } from '@ember/service';
 import { serviceLinks } from 'ember-osf/const/service-links';
+import { task } from 'ember-concurrency';
 
 export default Component.extend({
     serviceLinks,
+    store: service(),
     classNames: ['content'],
+
+    init() {
```
laurenbarker: It makes testing a little easier if you use `didReceiveAttrs` instead of `init`.

app/components/notification-list/template.hbs
```diff
@@ -1,19 +1,27 @@
...
 
-            <div class="notification-list-body">
-                {{#each subscriptions as |subscription|}}
-                    <hr>
-                    {{notification-list-row subscription=subscription}}
-                    <hr>
-                {{/each}}
-            </div>
+            {{#if fetchData.isRunning}}
+                {{loading-indicator classes='ball-dark text-center'}}
```
laurenbarker: Would you mind using a skeleton screen instead of a the ball scale loading here? We are in the process of phasing those loaders out.

app/components/notification-list/template.hbs
```diff
@@ -0,0 +1,27 @@
...
+                <span class="helper-text">
+                    {{t 'components.notificationList.helperText'}}
+                    <a href={{serviceLinks.settings}}>
+                        {{t 'components.notificationList.userSettings'}}
+                    </a>.
+                </span>
+            </div>
+
+            {{#if fetchData.isRunning}}
+                {{loading-indicator classes='ball-dark text-center'}}
```
laurenbarker: Would you mind using a skeleton screen instead of a the ball scale loading here? We are in the process of phasing those loaders out.

app/routes/preprints/provider/notifications.js
```diff
@@ -0,0 +1,15 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default Route.extend({
+    theme: service(),
+
+    model() {
+        const providerId = this.get('theme.id');
+        const subscriptionId = `${providerId}_new_pending_submissions`;
```
laurenbarker: Is there a benefit to assigning these in the model hook instead of in the controller?

app/components/notification-list-row/template.hbs
```diff
@@ -0,0 +1,19 @@
+<div class="flex-container">
+    {{#if updateFrequency.isRunning}}
+        {{loading-indicator classes='ball-dark text-center'}}
```
laurenbarker: Would you mind using a skeleton screen instead of a the ball scale loading here as well? We are in the process of phasing those loaders out.

app/components/notification-list-row/component.js
```diff
@@ -0,0 +1,43 @@
...
+        const frequency = this.get('subscription.frequency');
+        return this.get('notificationOptions').findBy('value', frequency).label;
+    }),
+
+    updateFrequency: task(function* (frequency) {
+        const subscription = this.get('subscription');
+        try {
+            subscription.set('frequency', frequency);
+            if (subscription.changedAttributes()) {
+                yield subscription.save();
```
laurenbarker, line 36: What happens if this yield never gets hit? Does the task keep running?

laurenbarker (changes requested):
Looking good! Just a couple more things

app/components/notification-list-row/component.js
adlius: I don't think so. `yield` is similar to `await` in Python, which, according to Ember Concurrency docs, "lets you pause execution of your task function until that promise resolves". I do not think `yield` is by any means necessary in a task. Therefore, if `yield` is not hit, the task would proceed without waiting for a promise.

app/components/moderation-base/component.js
```diff
@@ -30,6 +30,10 @@ export default Component.extend({
                 nameKey: 'global.settings',
                 route: 'preprints.provider.settings',
             },
+            {
```
laurenbarker: As per the mockups, the "Notifications" tab should be before the "Settings" tab.
screen shot 2018-04-02 at 1 30 48 pm

app/components/provider-links/template.hbs
```diff
@@ -28,6 +28,11 @@
                     {{t 'global.settings'}}
                 {{/link-to}}
             </li>
+            <li>
```
laurenbarker: "Notifications" should also be above "Settings" here since the "Notifications" tab comes first on the list view.

app/components/reviews-breadcrumbs/style.scss
```diff
@@ -4,6 +4,7 @@ a {
 
 .current-page {
   font-weight: bolder;
+  text-transform: capitalize;
```
laurenbarker: Do we still need this since the tab name doesn't appear on the breadcrumbs?

app/locales/en/translations.js
```diff
@@ -230,6 +231,17 @@ export default {
                 rejected: 'Rejected {{timeDate}} by {{moderatorName}}',
             },
         },
+        notificationList: {
+            heading: 'Configure reviews notification preferences',
+            helperText: 'To configure other notification preferences please visit your',
```
laurenbarker: Was "please" added to the language here during demo?

app/components/notification-list/template.hbs
```diff
@@ -0,0 +1,27 @@
+<div class="container">
+    <div class="row">
+        <div class="notification-list col-md-8">
+            <div class="notification-list-heading-container">
+                <h4 class="notification-list-heading">{{t 'components.notificationList.heading'}}</h4>
+                <span class="helper-text">
+                    {{t 'components.notificationList.helperText'}}
+                    <a href={{serviceLinks.settings}}>
```
laurenbarker: Can we have this go to `/settings/notifications/` instead of `/settings/`?

app/components/notification-list-row/template.hbs
```diff
@@ -0,0 +1,19 @@
+{{#if updateFrequency.isRunning}}
+    {{notification-list-row-skeleton}}
+{{else}}
+    <div class="flex-container">
+        <div class="notification-title">New pending submissions</div>
+        {{#bs-dropdown as |dd|}}
+            {{#dd.button class="dropdown"}} {{t currentSettingLabel}} <span class="fa fa-caret-down p-l-xs"></span>{{/dd.button}}
+            {{#dd.menu classNames="notification-menu" as |menu|}}
+                {{#each notificationOptions as |notificationOption| }}
```
laurenbarker: When hovering over items in the menu the row should be highlighted.
Expected:
screen shot 2018-04-02 at 1 44 01 pm

Actual:
screen shot 2018-04-02 at 1 42 12 pm

app/components/notification-list-row-skeleton/template.hbs
```diff
@@ -0,0 +1,5 @@
+{{#content-placeholders as |placeholder|}}
+    <div class="notification-title--skeleton">
```
laurenbarker: You could leave the top line out of the if/else check and add the bottom line of the row to the skeleton to make it look a little more like what will appear.

screen shot 2018-04-02 at 1 46 44 pm

app/components/notification-list-row/template.hbs
```diff
@@ -0,0 +1,19 @@
+{{#if updateFrequency.isRunning}}
```
laurenbarker: While `updateFrequency` is running you don't need to show a skeleton. It would probably make the most sense to just add `.enqueue()` to the task (see [docs](http://ember-concurrency.com/docs/task-concurrency) for details) so that if the user changes their selection multiple times the tasks won't interfere with one another.

laurenbarker: Since you don't need the skeleton here you can just put the skeleton in the `notification-list` template instead of having a component.

app/components/notification-list/template.hbs
```diff
@@ -0,0 +1,27 @@
+<div class="container">
+    <div class="row">
+        <div class="notification-list col-md-8">
+            <div class="notification-list-heading-container">
+                <h4 class="notification-list-heading">{{t 'components.notificationList.heading'}}</h4>
+                <span class="helper-text">
```
laurenbarker: If this is a `div` instead of a `span` it will keep it's padding when wrapping:
screen shot 2018-04-02 at 2 03 42 pm

laurenbarker (changes requested):
Checked it out and tested. Works well -- just a few easy changes and this should be GTM 🥇

app/components/notification-list/template.hbs
```diff
@@ -0,0 +1,32 @@
...
+                    <a href={{serviceLinks.settingsNotifications}}>
+                        {{t 'components.notificationList.userSettings'}}
+                    </a>.
+                </div>
+            </div>
+
+            <div class="line-separator"></div>
+            {{#if fetchData.isRunning}}
+                {{#content-placeholders as |placeholder|}}
+                    <div class="notification-title--skeleton">
```
laurenbarker, line 17: The placeholder isn't centered within the row (partly because there is a margin):
screen shot 2018-04-04 at 10 01 16 am

screen shot 2018-04-04 at 10 00 49 am

app/components/notification-list/template.hbs
```diff
@@ -0,0 +1,32 @@
+<div class="container">
+    <div class="row">
+        <div class="notification-list col-md-8">
+            <div class="notification-list-heading-container">
+                <h4 class="notification-list-heading">{{t 'components.notificationList.heading'}}</h4>
+                <div class="helper-text">
+                    {{t 'components.notificationList.helperText'}}
+                    <a href={{serviceLinks.settingsNotifications}}>
+                        {{t 'components.notificationList.userSettings'}}
```
laurenbarker: There is some extra space after the link here. Add a tilde to remove it: `{{t 'components.notificationList.userSettings'~}}`.

Before:
screen shot 2018-04-04 at 9 58 45 am

After:
screen shot 2018-04-04 at 10 14 42 am


See [stack overflow](https://stackoverflow.com/questions/15953293/handlebars-whitespace-control) or [Handlebars docs](http://handlebarsjs.com/expressions.html) for more info on controlling whitespace in handlebars.

app/components/notification-list-row/template.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="flex-container">
+    <div class="notification-title">New pending submissions</div>
+    {{#bs-dropdown as |dd|}}
+        {{#dd.button class="dropdown"}} {{t currentSettingLabel}} <span class="fa fa-caret-down p-l-xs"></span>{{/dd.button}}
+        {{#dd.menu classNames="notification-menu" as |menu|}}
+            {{#each notificationOptions as |notificationOption| }}
+                {{#menu.item}}
+                    <button class="btn" onclick={{perform updateFrequency notificationOption.value}}>
```
laurenbarker, line 8: Can we give this button a `min-width` of 86px so that the button size doesn't change when selecting different options. It will matter more when we add other settings to the page but it's better to take care of it now.

Then to make it so the spacing is dynamic between the label and the caret add:
```
.notification-menu {
  button {
    display: flex;
    justify-content: space-between;
  }
}
```

laurenbarker (changes requested):
Alright this should be the last back and forth. Just a few more nitpicks 🌈

app/components/notification-list-row/template.hbs
adlius: I believe you are referring to the button 4 lines above. (See the previous comment.) If so, I've added a new css file for the `notification-list-row` component to address the width and spacing issue.

app/components/notification-list-row/template.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="flex-container">
+    <div class="notification-title">New pending submissions</div>
+    {{#bs-dropdown as |dd|}}
+        {{#dd.button class="dropdown"}} {{t currentSettingLabel}} <span class="fa fa-caret-down p-l-xs"></span>{{/dd.button}}
```
adlius, line 4: @laurenbarker I believe you are referring to this button.

laurenbarker: Yup, that is the button I meant!

laurenbarker (approved):

app/components/notification-list/style.scss
```diff
@@ -0,0 +1,51 @@
+.notification-list-heading-container{
+  padding-bottom: 10px;
+}
+.notification-list-heading {
+  padding: 20px 20px 0px 20px;
```
alexschiller: very minor, the 0 doesn't need `px`

app/components/notification-list/style.scss
```diff
@@ -0,0 +1,51 @@
...
+  margin-top: 9px;
+  margin-bottom: 9px;
+}
+
+.line-separator {
+  border-bottom: 1px solid $color-bg-gray-light;
+}
+
+.flex-container {
+  padding-top: 20px;
```
alexschiller: this hits each, can you just use `padding: 20px;`

alexschiller (changes requested):
Added two minor comments that aren't worth another commit.

Please add QA notes (thus requesting changes and not approved).

alexschiller (approved):

PR #135, opened 2018-05-21
[IN-6][Reviews] Moderator management fixes

laurenbarker:
## Purpose
Improvements and bug fixes found by QA.

## Changes
* Increase page size for admin moderator request to make sure that we get all of the admins
    * this is a temporary solution until we have a solution for nested routes in general
* Make toast error more informative if trying to add a user who is already a moderator
* Scroll to form when adding a new moderator so the user is sure to see form
* Handle when user removes self as moderator
* Prevent tabs from wrapping on small screens by reducing font size/padding
screen shot 2018-05-18 at 3 30 47 pm
screen shot 2018-05-18 at 3 30 56 pm
screen shot 2018-05-18 at 3 31 04 pm

## QA notes
Everything is a fix for something that was reported on the ticket. Would spend a little more time on the scroll functionality -- would be good to test how it looks when clicking the add button on the first page and then going to a second page that is long.

## Ticket


app/controllers/preprints/provider/moderators.js
```diff
@@ -81,10 +83,15 @@ export default Controller.extend(Analytics, moderatorsQueryParams.Mixin, {
     }),
 
     deleteModerator: task(function* (moderatorInstance) {
+        const removingSelf = moderatorInstance.id === this.get('currentUser.user.id');
         try {
             yield moderatorInstance.destroyRecord({ adapterOptions: { provider: this.get('theme.provider.id') } });
             moderatorInstance.unloadRecord(); // https://github.com/emberjs/data/issues/5014
 
+            if (removingSelf) {
+                return;
```
alexschiller, line 92: this just `returns` but the others return `true` or `false`... but it kinda looks like the `true` and `false` might be unused?

app/controllers/preprints/provider/moderators.js
```diff
@@ -81,10 +83,15 @@ export default Controller.extend(Analytics, moderatorsQueryParams.Mixin, {
     }),
 
     deleteModerator: task(function* (moderatorInstance) {
+        const removingSelf = moderatorInstance.id === this.get('currentUser.user.id');
         try {
```
alexschiller: What part in this `try` is the part that might be expected to fail? It looks like if the if the destroyRecord/unload fails on removing the self, then the user will still be redirected to `/` which might be unexpected if the delete doesn't work. 

As an aside, `return`s in `try-catch-finally` do not behave how I would expect coming from python 😖

I don't think this necessarily needs to be changed but I wanted to check with you before merging.

alexschiller (changes requested):
Two quick comments otherwise LGTM... let me know if you think it should be changed otherwise I'll 🚢

app/controllers/preprints/provider/moderators.js
laurenbarker: Ah the other returns aren't actually needed -- I think the need was refactored away. This return is so none of the other code is run except the finally.

app/controllers/preprints/provider/moderators.js
laurenbarker: @alexschiller that's a good point. Can you think of a better way than just adding another variable that gets set if it makes it past the destroyRecord part without error?

app/controllers/preprints/provider/moderators.js
alexschiller: @laurenbarker A little hacky, but I think setting removingSelf to false in the `catch` would do the trick. Would add a comment to make it clear why. Alternative is two separate workflows... which seems unnecessary. 

Edit: For reference, this comment was meant to go as a response to the destroy record variable question. Woops.

alexschiller (approved):
LGTM 🌾 🌴

PR #136, opened 2018-05-21
[IN-6][Reviews] Fix scroll to add form

laurenbarker:
## Purpose
Fix scroll to add form.

## Changes
* Use reliable class name to select last row

## Ticket


alexschiller (approved):
👍

PR #138, opened 2018-06-04
[IN-254][Reviews] Update preprint detail to use name instead of given/full

baylee-d:
## Purpose
In some cases contributors might use the `fullName` field instead of `familyName` and `givenName`.  On the preprint detail page for moderated providers in the reviews app, we are currently only looking at given/family names to make the contributor name.  A better solution to this would be to use the `name` field, which combines given/family, or if not seeing either of those uses fullName.

## Changes
- Use `name` instead of `familyName` and `givenName` fields
- Update `contributor-list` test

## QA
Test this by:

- Editing a user to only be using `fullName` in their settings.  Make sure given/family/middle name fields are blank
- Create a preprint using a moderated service
- View the preprint.  Make sure that the user's name is showing up under the title and that it matches what you entered for `fullName`

You can also test this using given/family/middle names instead of `fullName` to make sure that those show up as the user's name if they are filled out.

## Screenshots
If given/family/middle name fields are filled out, it should use those as the placeholder on the detail page:
screen shot 2018-06-04 at 11 09 07 am
screen shot 2018-06-04 at 11 09 32 am

But if they aren't filled out and there's a `fullName` entered, that should show up instead:
screen shot 2018-06-04 at 11 11 42 am
screen shot 2018-06-04 at 11 10 02 am

## Ticket
https://openscience.atlassian.net/browse/IN-254


alexschiller (approved):
LGTM

PR #139, opened 2018-06-07
[IN-298][Reviews] Prevent banner load with undefined preprint

baylee-d:
## Purpose

Currently the banner will load on preprints detail page, even if the preprint is not defined.  This results in a missing translation error.

## Changes

- Add conditional to the banner to only display if the preprint is not loading or if the preprint is defined

## QA

Changes will be seen on staging2.  If the preprint is undefined, the page should look like this:
screen shot 2018-06-07 at 1 11 54 pm

But if there is a preprint, it should still show up like normal:
screen shot 2018-06-07 at 2 03 12 pm

## Ticket

https://openscience.atlassian.net/browse/IN-298


alexschiller (approved):
LGTM, changelog looks weird due to diff issues I've seen a few times. I will check it after merging to make sure it is in the right place.

PR #140, opened 2018-06-07
[IN-219][Reviews] Ember-Component-CSS -> Ember-CSS-Modules

chrisseto:
Switch from Ember-Component-CSS to Ember-CSS-Modules as Ember-Component-CSS does not support Ember-Engines.

QA Notes:
Almost every style sheet was touched in this PR, but there should be no noticeable changes.
The one exception being that the status headers on the submissions page will no longer become bold if you right click on them. (See the below screenshot)

screen shot 2018-06-07 at 3 41 10 pm


app/components/notification-list-row/styles.scss
```diff
@@ -0,0 +1,20 @@
+.dropdown {
+    min-width: 86px;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+}
+
+.flex-container {
+    // align-items: center;
```
laurenbarker: Delete?

ember-cli-build.js
```diff
@@ -1,12 +1,9 @@
 /* eslint-env node */
-
-
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 const Funnel = require('broccoli-funnel');
 const jQueryPackage = require('jquery/package.json');
-const Autoprefixer = require('autoprefixer');
-const CSSNano = require('cssnano');
-
+// const Autoprefixer = require('autoprefixer');
```
laurenbarker: Delete?

ember-cli-build.js
```diff
@@ -81,20 +78,31 @@ module.exports = function(defaults) {
...
-                }, {
-                    module: CSSNano,
-                }],
-            },
+        cssModules: {
+            headerModules: [
+                'reviews/styles/headers',
+            ],
         },
+        // cssModules: {
```
laurenbarker: Also, delete?

laurenbarker (dismissed):
Could you add a PR description with QA notes? 

Other than the random commented out stuff and the lack of a description this looks great. Thanks for also adding style linting! 

Were there styles you ended up actually changing?

chrisseto:
I cleaned up some of the styles but everything should look the same.

chrisseto:
side by side: http://recordit.co/OD7omLPFr6
Looking into the moderators tab and preprint preview page oddities.

laurenbarker:
The skeleton screens for the provider section on the dashboard got stretched.
Before:
providers-normal

After:
providers-stretched

chrisseto:
Nice catch!

app/components/action-feed-entry/styles.scss
```diff
@@ -8,6 +8,9 @@
 
 }
 
+// Used in Unit tests
+.action-body { } /* stylelint-disable-line block-no-empty */
```
laurenbarker: Why does the class need to be in the stylesheet? Do unused styles get removed on compile or something?

chrisseto: css modules munges all the class names when the app is built. It's in the style sheet so the tests can get the "compiled" name. I could just leave it untouched, as a `class` rather than `local-class` but it just looks a bit odd. Or do a weird selector. Whichever works for me.

laurenbarker: I don't really have a preference. I'll leave that up to @alexschiller

alexschiller:
Looked everything over. Fixed merge conflict. Checked out PR locally and everything looked good. Will wait for TCI to be green after the conflict fix and then get it merged.

alexschiller (approved):
wew

chrisseto:
Thanks @alexschiller

PR #143, opened 2018-06-20
[IN-338][Reviews] Bump eosf version to fix yarn

alexschiller:
yarn is bork. fix


fabmiz (approved):
yarn lockfile updated. Looks pretty good 🥇

